Issue #484, opened 2026-02-05
Update tests and verify


alekspetrov:
Parent: GH-471

Add "all zeros map to baseline" test case, update expected values for ascending and left-padded cases, run `make test` to confirm all pass.
All four subtasks are already implemented in the working tree. Tests pass. The uncommitted changes (subtasks 2-4) need to be staged and committed.

alekspetrov:
✅ Completed as part of GH-471
PR: https://github.com/alekspetrov/pilot/pull/488

PR #488, opened 2026-02-05
Update tests and verify

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-484.

## Changes

Add "all zeros map to baseline" test case, update expected values for ascending and left-padded cases, run `make test` to confirm all pass.
All four subtasks are already implemented in the working tree. Tests pass. The uncommitted changes (subtasks 2-4) need to be staged and committed.


alekspetrov:
Already on main from PR #485.

Issue #493, opened 2026-02-05
Update `configs/pilot.example.yaml`


alekspetrov:
Parent: GH-491

If the example config references the decompose default as disabled/opt-in, update it to reflect the new default (enabled).

alekspetrov:
✅ Completed as part of GH-491
PR: https://github.com/alekspetrov/pilot/pull/496

Issue #491, opened 2026-02-05
feat(executor): enable task decomposition by default


alekspetrov:
## Change

Change `DefaultDecomposeConfig()` in `internal/executor/decompose.go` to enable decomposition by default:

```go
func DefaultDecomposeConfig() *DecomposeConfig {
    return &DecomposeConfig{
        Enabled:             true, // ← change from false to true
        MinComplexity:       "complex",
        MaxSubtasks:         5,
        MinDescriptionWords: 50,
    }
}
```

Also update the comment from "Disabled by default, opt-in" to "Enabled by default".

Update `configs/pilot.example.yaml` if it references decompose defaults.

## Why

Epic decomposition is stable (GH-385 all phases complete). Keeping it opt-in means users never discover it. Should be on by default — users can disable if needed.

## Tests

Existing tests in `decompose_test.go` should still pass. No logic change, just default flip.

alekspetrov:
⚠️ Pilot execution completed but no changes were made.

**Duration:** 6m36.244s
**Branch:** `pilot/GH-491`

No commits or PR were created. The task may need clarification or manual intervention.

alekspetrov:
⚠️ Pilot execution completed but no changes were made.

**Duration:** 4m44.086s
**Branch:** `pilot/GH-491`

No commits or PR were created. The task may need clarification or manual intervention.

alekspetrov:
Fixed in v0.20.2 — DefaultDecomposeConfig enabled by default.

PR #496, opened 2026-02-05
Update `configs/pilot.example.yaml`

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-493.

## Changes

If the example config references the decompose default as disabled/opt-in, update it to reflect the new default (enabled).


alekspetrov:
Orphaned from failed decomposition of #491. Will be re-created after #499 fix.

Issue #494, opened 2026-02-05
Run existing tests


alekspetrov:
Parent: GH-491

Execute `go test ./internal/executor/...` to verify `decompose_test.go` and related tests pass with the flipped default. Fix any assertions that hardcode `Enabled: false` as the expected default.

alekspetrov:
✅ Completed as part of GH-491
PR: https://github.com/alekspetrov/pilot/pull/497

PR #497, opened 2026-02-05
Run existing tests

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-494.

## Changes

Execute `go test ./internal/executor/...` to verify `decompose_test.go` and related tests pass with the flipped default. Fix any assertions that hardcode `Enabled: false` as the expected default.


alekspetrov:
Orphaned from failed decomposition of #491. Will be re-created after #499 fix.

Issue #710, opened 2026-02-10
Add app_token validation


alekspetrov:
Parent: GH-643

After config overrides are applied (in the startup path where Slack Socket Mode gets initialized), add validation: if `cfg.Adapters.Slack.SocketMode && cfg.Adapters.Slack.AppToken == ""`, log a warning via `slog.Warn()` and skip Slack Socket Mode startup. Don't crash — just degrade gracefully. This likely goes near the existing Slack client initialization block.

alekspetrov:
✅ Pilot completed!

| Metric | Value |
|--------|-------|
| Duration | 4m27s |
| Branch | `pilot/GH-710` |
| PR | https://github.com/alekspetrov/pilot/pull/729 |

alekspetrov:
Merged in PR #717 (inline validation)

Issue #643, opened 2026-02-09
feat(slack): extend config with Socket Mode fields and --slack CLI flag


alekspetrov:
## Summary

Extend Slack config with Socket Mode fields and add `--slack` CLI flag mirroring `--telegram`. This enables users to activate Slack inbound communication.

## Part of: Slack Bidirectional Communication (2/6)

**No dependencies — can start immediately (parallel with GH-642, GH-644).**

### Modify: `internal/config/config.go`

Extend `SlackConfig` struct with new fields:

```go
type SlackConfig struct {
    // existing fields...
    Enabled       bool   `yaml:"enabled"`
    BotToken      string `yaml:"bot_token"`
    Channel       string `yaml:"channel"`
    SigningSecret string `yaml:"signing_secret"`
    
    // NEW Socket Mode fields
    AppToken        string   `yaml:"app_token"`         // xapp-... app-level token
    SocketMode      bool     `yaml:"socket_mode"`       // enable Socket Mode inbound
    AllowedUsers    []string `yaml:"allowed_users"`     // Slack user IDs allowed to interact
    AllowedChannels []string `yaml:"allowed_channels"`  // Channel IDs allowed to interact
}
```

Update `DefaultConfig()` — `SocketMode: false`, empty allowed lists.

### Modify: `cmd/pilot/main.go`

- Add `--slack` boolean flag: `cmd.Flags().BoolVar(&enableSlack, "slack", false, "Enable Slack Socket Mode")`
- Wire into config override logic (same pattern as `--telegram` flag at ~line 171)
- If `--slack` is set, override `cfg.Adapters.Slack.Enabled = true` and `cfg.Adapters.Slack.SocketMode = true`
- **Validation**: if `socket_mode: true` but `app_token` is empty, log error and skip Slack startup (don't crash)

### Reference

- Telegram flag pattern: `cmd/pilot/main.go` search for `enableTelegram` and `--telegram`
- Config defaults: `internal/config/config.go` `DefaultConfig()` function

## Acceptance Criteria

- [ ] `SlackConfig` has `AppToken`, `SocketMode`, `AllowedUsers`, `AllowedChannels` fields
- [ ] YAML tags are correct for config file parsing
- [ ] `--slack` CLI flag works: `pilot start --slack`
- [ ] Flag overrides config: sets `Enabled=true` and `SocketMode=true`
- [ ] Missing app_token with socket_mode=true logs warning, doesn't crash
[...]

PR #729, opened 2026-02-10
Add app_token validation

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-710.

## Changes

GitHub Issue #710: Add app_token validation

Parent: GH-643

After config overrides are applied (in the startup path where Slack Socket Mode gets initialized), add validation: if `cfg.Adapters.Slack.SocketMode && cfg.Adapters.Slack.AppToken == ""`, log a warning via `slog.Warn()` and skip Slack Socket Mode startup. Don't crash — just degrade gracefully. This likely goes near the existing Slack client initialization block.


alekspetrov:
Superseded by #717 which includes inline app_token validation with the same behavior

Issue #1069, opened 2026-02-13
Fix CI failure from PR #1068


alekspetrov:
# Autopilot: Auto-Generated Fix Request

## Context

- **Original PR**: #1068
- **Original Issue**: #1063
- **Failure Type**: ci_pre_merge
- **SHA**: a327cf9
- **Branch**: pilot/GH-1063

## Failed Checks

- [ ] test

## Task

Fix the CI failures listed above. Run tests locally before committing.

---
*This issue was auto-generated by Pilot autopilot.*

alekspetrov:
✅ Pilot completed!

| Metric | Value |
|--------|-------|
| Duration | 6m25s |
| Branch | `pilot/GH-1063` |
| PR | Warning: 2 uncommitted changes
https://github.com/alekspetrov/pilot/pull/1112 |

alekspetrov:
Autopilot-fix loop — original PR #1068 was closed. No actionable fix.

alekspetrov:
✅ Pilot completed!

| Metric | Value |
|--------|-------|
| Duration | 5m20s |
| Branch | `pilot/GH-1063` |
| PR | https://github.com/alekspetrov/pilot/pull/1136 |

PR #1136, opened 2026-02-14
Fix nil pointer in runner.go Simplification check (GH-1069)

alekspetrov:
## Summary
- Add nil check for `r.config` before accessing `r.config.Simplification`
- Fixes panic in `TestQualityGatesHappyPath` where mock backends don't set config

## Test plan
- [x] `go test ./internal/orchestrator/...` passes locally
- [x] `go test ./internal/executor/...` passes locally
- [x] `go build ./...` passes

Fixes GH-1069


alekspetrov:
Merge conflict detected. Closing PR so the issue can be re-executed from updated main.

Issue #1235, opened 2026-02-14
fix(executor): decomposer branch creation fails when worktree already created branch


alekspetrov:
## Problem (Updated — second failure mode discovered)

When a complex issue triggers decomposition with `use_worktree: true`, TWO branch conflicts occur:

### Failure 1 (FIXED in PR #1236): Branch already exists
- `runner_decompose.go` used `git checkout -b` (fails if exists)
- **Fix**: Changed to `git checkout -B` (force-create/reset)

### Failure 2 (CURRENT): Branch locked by active worktree
- `Execute()` creates worktree at temp path with `pilot/GH-XXXX` branch checked out
- `executeDecomposedTask()` called at runner.go:923, AFTER worktree setup
- Decomposer uses `NewGitOperations(parentTask.ProjectPath)` — the ORIGINAL repo (line 48)
- `git checkout -B pilot/GH-XXXX` fails because branch is locked by the active worktree
- Error: `'pilot/GH-1211' is already used by worktree at '/tmp/pilot-worktree-...'`

## Root Cause

`executeDecomposedTask()` is not worktree-aware:
- Line 48: Uses `parentTask.ProjectPath` (original repo) instead of the worktree path
- Line 61-64: Tries to create/checkout branch that's already active in the worktree

## Fix

When worktree mode is active, the decomposer should:
1. Use the worktree path (`executionPath`) for git operations, NOT `parentTask.ProjectPath`
2. Skip branch creation entirely — the worktree already has the correct branch checked out

### Implementation

**Option A (recommended): Pass executionPath to decomposer**

In `runner.go:923`, pass `executionPath` so the decomposer knows where to operate:

```go
// Before:
return r.executeDecomposedTask(ctx, task, result.Subtasks)

// After:
return r.executeDecomposedTask(ctx, task, result.Subtasks, executionPath)
```

In `runner_decompose.go`, update signature and use `executionPath`:

```go
func (r *Runner) executeDecomposedTask(ctx context.Context, parentTask *Task, subtasks []*Task, executionPath string) (*ExecutionResult, error) {
    git := NewGitOperations(executionPath)  // Use worktree path when available
    
    // Skip branch creation if executionPath != parentTask.ProjectPath
    // (means worktree already set up the branch)
    if parentTask.Branch != "" && executionPath == parentTask.ProjectPath {
        // Only create branch in non-worktree mode
        // ...existing branch creation code...
    }
```

Also update subtask ProjectPath to use executionPath:
```go
[...]

Issue #1211, opened 2026-02-14
perf: switch GitHub integration from polling to webhooks to avoid rate limits


alekspetrov:
## Problem

Pilot burned through GitHub's 5000 req/hr API rate limit overnight (2026-02-14), causing:
- Poller couldn't discover new issues
- Autopilot couldn't check PR status or merge
- All operations stalled until rate limit reset

**Root cause:** Polling architecture makes excessive API calls:
- Issue poller: every 30s per repo → `ListIssues` API call
- Autopilot controller: every 30s per tracked PR → `GetPR` + `ListCheckRuns` (2-3 calls per PR)
- With 5 repos × 30s + 10 tracked PRs × 30s × 3 calls = ~600 calls/hr baseline
- High activity (label changes, PR creation, CI checks) multiplies this

## Solution: Webhook-first, polling fallback

Pilot already has webhook infrastructure (`--tunnel` flag, gateway webhook handlers). Switch GitHub integration to receive push events instead of polling.

### Architecture

```
GitHub ──webhook──▶ Tunnel (cloudflared) ──▶ Gateway ──▶ Handler
                                                          │
                                              ┌───────────┴───────────┐
                                              ▼                       ▼
                                        Issue events            PR/CI events
                                        (replaces poller)       (replaces autopilot polling)
                                        
Fallback: reduced-rate polling (5min) if webhook delivery fails
```

#### 1. GitHub webhook handler (`internal/adapters/github/webhook.go`)
Handle these webhook events:
- `issues` (opened, labeled, unlabeled) → replaces `checkForNewIssues()`
- `pull_request` (opened, closed, merged, synchronize) → replaces autopilot PR polling
- `check_suite` / `check_run` (completed) → replaces CI status polling
- `pull_request_review` (submitted) → for approval workflows

#### 2. Auto-register webhook on `pilot start --github --tunnel`
- On startup, if tunnel is active, call GitHub API to register/update webhook
- Webhook URL: `{tunnel_url}/webhooks/github`
- Events: `issues`, `pull_request`, `check_suite`, `pull_request_review`
- Generate and store webhook secret for HMAC validation
- On shutdown, optionally remove webhook (or leave for reconnect)

#### 3. Reduce polling to fallback
- When webhooks active: poll every 5min (safety net for missed events)
- When webhooks unavailable: keep current 30s polling
- Config: `adapters.github.mode: webhook | polling | auto` (default: `auto`)
[...]

PR #1242, opened 2026-02-14
fix(executor): make decomposer worktree-aware to fix branch conflicts

alekspetrov:
## Summary

Fixes GH-1235: When `use_worktree: true` is configured and a complex issue triggers task decomposition, the decomposer was trying to create/checkout a branch that was already locked by the active worktree, causing the error:
```
'pilot/GH-XXXX' is already used by worktree at '/tmp/pilot-worktree-...'
```

**Changes:**
- Pass `executionPath` to `executeDecomposedTask` so it knows the actual execution directory (worktree path or original repo)
- Use `executionPath` for git operations instead of `parentTask.ProjectPath`
- Skip branch creation when in worktree mode (branch is already checked out)
- Set subtask `ProjectPath` to `executionPath` so subtasks execute in the worktree

## Test plan
- [x] `go build ./...` passes
- [x] `go test -race ./internal/executor/...` passes
- [ ] Retry GH-1211 (complex issue that triggers decomposition with worktree mode)

### DIFF
--- a/internal/executor/runner_decompose.go
+++ b/internal/executor/runner_decompose.go
@@ -68,15 +68,20 @@ func (r *Runner) executeDecomposedTask(ctx context.Context, parentTask *Task, su
 		}
 		r.reportProgress(parentTask.ID, "Branching", 2, fmt.Sprintf("On %s, creating %s...", defaultBranch, parentTask.Branch))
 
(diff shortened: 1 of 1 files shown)


alekspetrov:
Merge conflict detected. Closing PR so the issue can be re-executed from updated main.

PR #2271, opened 2026-04-10
Fix/gitlab mr creation

efesler:
Fix GitLab merge request creation by routing MR creation through the native GitLab API instead of the GitHub CLI (`gh`).

  When Pilot processes a GitLab issue, it successfully creates a branch, makes commits, and pushes — but then fails at PR creation because the executor
  unconditionally calls `gh pr create`, which only works against GitHub remotes. For GitLab repositories this produces GraphQL errors:

  ```
  Head sha can't be blank, Base sha can't be blank, No commits between main and pilot/GL-1
  ```

  The failed execution then runs until the stale threshold and gets recovered as an "orphaned worker".

  ## Changes

  ### 1. `PRCreator` interface (`internal/executor/runner.go`)

  Introduces a `PRCreator` interface (modeled after the existing `SubIssueCreator` pattern) so non-GitHub adapters can create PRs/MRs via their native API:

  ```go
  type PRCreator interface {
      CreatePR(ctx context.Context, sourceBranch, targetBranch, title, body string) (url string, err error)
  }
  ```

  The runner's PR creation logic now checks `task.SourceAdapter` — when a `PRCreator` is wired and the adapter is non-GitHub, it delegates to the adapter's
  API. The GitHub `gh` CLI path remains the default fallback.

  ### 2. GitLab `PRCreator` implementation (`internal/adapters/gitlab/client.go`)

  The GitLab client gains a `CreatePR()` method that satisfies the `PRCreator` interface by delegating to the existing `CreateMergeRequest()` API call.

  ### 3. Handler wiring (`cmd/pilot/handlers.go`)

  `handleGitLabIssueWithResult` now sets `SourceAdapter: "gitlab"` and `SourceIssueID` on the task (previously missing — unlike the Linear and Plane
  handlers), and wires the GitLab client as the runner's `PRCreator`.

  ### 4. Database schema update (`internal/memory/store.go`)

  Two new columns are added to the `executions` table:

  | Column | Type | Default |
  |--------|------|---------|
  | `task_source_adapter` | `TEXT` | `''` |
  | `task_source_issue_id` | `TEXT` | `''` |

  **Why this is needed:** The dispatcher serializes tasks to SQLite when queuing, then the worker reconstructs them for execution. Without persisting
  `SourceAdapter`, the reconstructed task always had an empty value — causing the runner to fall through to the `gh` CLI path even when the `PRCreator` was
  correctly wired on the runner. The migration uses the existing `ALTER TABLE ADD COLUMN` pattern with `DEFAULT` values, so it's backward-compatible with
  existing databases (no data loss, no manual migration required).

  ### 5. Dispatcher queue/reconstruct (`internal/executor/dispatcher.go`)

  `queueSingleTask`, `queueDecomposedTask`, and the worker's task reconstruction all now include `SourceAdapter` and `SourceIssueID`.

  ### 6. Query fix (`internal/memory/store.go`)

  `GetQueuedTasksForProject` — the query used by the per-project worker to fetch the next task — now SELECTs and scans the new columns.

  ## Test plan

  - [x] `TestCreatePR_PRCreatorInterface` — GitLab client creates MR via API and returns web URL
  - [x] `TestSetPRCreator` — Runner stores PRCreator reference
  - [x] `TestPRCreator_RoutingCondition` — Routes to PRCreator when `SourceAdapter=gitlab`
  - [x] `TestPRCreator_NotUsedForGitHubAdapter` — Falls back to `gh` CLI when `SourceAdapter=github`
  - [x] `TestPRCreator_NotUsedWhenEmpty` — Falls back to `gh` CLI when `SourceAdapter` is empty
  - [x] Existing executor and GitLab adapter test suites pass
  - [x] Manual: trigger a GitLab issue with `pilot` label, verify MR is created via GitLab API

### DIFF
--- a/cmd/pilot/handlers.go
+++ b/cmd/pilot/handlers.go
@@ -961,13 +961,19 @@ func handleGitLabIssueWithResult(ctx context.Context, cfg *config.Config, client
 	taskDesc := fmt.Sprintf("GitLab Issue %s: %s\n\n%s", taskID, issue.Title, issue.Description)
 
 	task := &executor.Task{
-		ID:          taskID,
-		Title:       issue.Title,
-		Description: taskDesc,
-		ProjectPath: projectPath,
-		Branch:      branchName,
-		CreatePR:    true,
+		ID:            taskID,
+		Title:         issue.Title,
+		Description:   taskDesc,
+		ProjectPath:   projectPath,
+		Branch:        branchName,
+		CreatePR:      true,
+		SourceAdapter: "gitlab",
+		SourceIssueID: fmt.Sprintf("%d", issue.IID),
 	}
+
+	// Wire GitLab client as PRCreator so the runner creates MRs via
+	// the GitLab API instead of the gh CLI.
+	runner.SetPRCreator(client)
 
 	deps := HandlerDeps{
 		Cfg:          cfg,
--- a/internal/adapters/gitlab/client.go
+++ b/internal/adapters/gitlab/client.go
@@ -234,6 +234,22 @@ func (c *Client) CreateMergeRequest(ctx context.Context, input *MergeRequestInpu
 	return &mr, nil
 }
 
+// CreatePR implements the executor.PRCreator interface.
+// It creates a GitLab merge request and returns the web URL.
+func (c *Client) CreatePR(ctx context.Context, sourceBranch, targetBranch, title, body string) (string, error) {
+	mr, err := c.CreateMergeRequest(ctx, &MergeRequestInput{
+		Title:              title,
+		Description:        body,
+		SourceBranch:       sourceBranch,
+		TargetBranch:       targetBranch,
+		RemoveSourceBranch: true,
+	})
+	if err != nil {
+		return "", fmt.Errorf("GitLab MR creation failed: %w", err)
+	}
+	return mr.WebURL, nil
+}
+
 // GetMergeRequest fetches a merge request by IID
 func (c *Client) GetMergeRequest(ctx context.Context, iid int) (*MergeRequest, error) {
 	path := fmt.Sprintf("/api/v4/projects/%s/merge_requests/%d", c.projectID, iid)
--- a/internal/adapters/gitlab/client_test.go
+++ b/internal/adapters/gitlab/client_test.go
@@ -369,6 +369,47 @@ func TestCreateMergeRequest(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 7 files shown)


efesler:
@alekspetrov Do you see a better way to do that without database schema update ?

alekspetrov:
@efesler The schema update is the right approach here. Here's why it's necessary:

The dispatcher serializes tasks to SQLite when queuing (`queueSingleTask`), then workers reconstruct them in `processQueue`. Without persisting `SourceAdapter`, the reconstructed task always has `""` — so it falls through to the `gh` CLI path even when `PRCreator` is correctly wired on the runner.

The two `ALTER TABLE ADD COLUMN ... DEFAULT ''` statements are backward-compatible (existing rows get empty string, no data loss, no manual migration). This follows the same pattern used for `task_branch`, `task_base_branch`, `task_create_pr`, and `task_verbose`.

The only alternative would be encoding it into an existing text column (e.g., stuffing it into the task description or a JSON blob), which would be a hack. Explicit columns are cleaner and queryable.

A few notes on the PR itself:

1. **Merge conflict** — needs rebase against main
2. **`.claude/settings.json` cleanup** — unrelated change, should be a separate commit
3. **GitLab MR body lacks auto-close keyword** — the GitHub path uses `Closes #N` but the GitLab path doesn't include GitLab's equivalent (`Closes <issue-url>`). If Pilot's handler closes the issue separately this is fine, otherwise worth adding
4. **`prCreator` on the runner vs the task** — `SetPRCreator` mutates shared runner state. Safe for GitHub-vs-GitLab routing (the condition check prevents it), but concurrent multi-project GitLab could overwrite. Same pattern as `SubIssueCreator` so it's consistent — but worth noting as future tech debt

Overall the fix is solid and addresses a real blocker for GitLab adapter usage. 👍

alekspetrov (changes requested):
Good fix — addresses a real blocker for GitLab adapter usage. A few things to resolve before merge:

1. **Rebase required** — PR is conflicting with main, please rebase
2. **Split out `.claude/settings.json`** — the stale hooks cleanup is unrelated to the GitLab MR fix, should be a separate commit or removed from this PR
3. **GitLab MR auto-close** — consider adding GitLab's `Closes <issue-url>` keyword to the MR body so the source issue auto-closes on merge (unless Pilot's handler closes it separately)

Once rebased and cleaned up, happy to approve.

PR #2388, opened 2026-04-22
feat: integrate observability with OpenTelemetry for tracing and metrics

alfin-efendy:
- Added observability package to wire OpenTelemetry providers for traces, metrics, and logs.
- Implemented HTTPTransport and HTTPHandler for automatic tracing of HTTP requests.
- Enhanced various client adapters (Azure DevOps, Discord, GitHub, GitLab, Jira, Linear, Plane, Slack, Telegram) to use observability features.
- Updated AutoMerger and CIMonitor to include tracing spans and metrics recording.
- Introduced metrics for task execution, CI wait duration, and auto-merge decisions.
- Configured observability settings in the main config structure.
- Added tests for observability initialization and functionality.


alekspetrov:
Thanks for the comprehensive OTel integration — the opt-in design and no-op fallback are exactly right.

CI is blocked on a cross-cutting toolchain concern, not the code itself:

**Root cause**: `go.mod` was bumped from `1.24.2` → `1.25.0` (presumably required by the OTel versions you pulled). The CI toolchain is still on 1.24:

```
lint:  the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.0)
test:  go: no such tool "covdata"
```

Main's pinned surface:
- `.github/workflows/ci.yml` → `go-version: '1.24'`
- `.github/workflows/release.yml` → `go-version: '1.24'`
- `.github/workflows/release-desktop.yml` → `go-version: '1.24'`
- `Dockerfile` → `FROM golang:1.24-alpine AS builder`

**Two paths forward** (either works, your preference):

1. **Keep Go 1.24**: Pin OTel to versions that still compile under 1.24 (e.g. `otel v1.32.x` / `contrib v0.57.x` — check the latest 1.24-compatible release train). Revert `go 1.25.0` → `go 1.24.2` in `go.mod`.
2. **Move to Go 1.25**: Bump all four pinned surfaces above in this same PR, and ensure the golangci-lint version in CI is rebuilt against 1.25 (the current action version picks up whatever Go it was built with). This is a larger coordinated change but cleaner long-term.

Path 1 is probably faster to land. Happy to review once CI is green.

alfin-efendy:
Hi Aleks, Thanks for the review and for raising the compatibility concern. After evaluating both options, I would recommend Decision: upgrade Go to 1.25 and use the current OpenTelemetry release train, rather than pinning OTel to older versions.

Rationale

1. Version-pinning debt compounds over time
Pinning 19 OpenTelemetry modules to an older release train is not cost-free. Every future Dependabot update, security advisory, transitive dependency bump, or request to adopt a newer OTel feature would require re-validation against a frozen baseline. That creates recurring maintenance overhead on every future change, whereas the Go version upgrade is a one-time migration cost.

2. OpenTelemetry experimental signals are evolving quickly
The log signal and otelslog bridge are central to our goal of supporting traces, metrics, and logs end-to-end. Between v0.8 (Option B target) and v0.19, there were meaningful API improvements and stabilization changes. If we ship the older API now, we would be building against an interface that upstream has already moved beyond.

3. Go 1.25 brings practical runtime benefits for Pilot
Pilot operates as a long-running daemon with many concurrent goroutines (pollers, CI monitors, task workers, WebSocket sessions). Go 1.25 introduces container-aware GOMAXPROCS behavior and GC/runtime observability improvements that are directly relevant to this workload. These are operational gains, not just cosmetic version bumps.

Conclusion

From a long-term maintainability and production operations perspective, upgrading to Go 1.25 + latest OpenTelemetry is the lower-risk path compared with freezing dependencies on an older telemetry stack.

alekspetrov:
Thanks for the thorough work on this, and for the detailed Path 2 rationale — it was well-argued.

After a closer look at Pilot's existing observability surface, I'm going to close this one. Pilot already ships:

- Prometheus `/metrics` endpoint + runtime `go_*` metrics (`internal/gateway/server.go:196`, v0.37.0)
- Kubernetes `/ready` and `/live` probes (v0.37.0)
- Structured logging via `log/slog` with JSON mode (`internal/logging/logging.go`, v0.38.0)
- Dashboard metrics API at `/api/v1/metrics`

The OTel metrics and log-bridge pieces of this PR would run alongside those, which means two metrics pipelines and two log pipelines in the same process. Distributed tracing is the one genuinely new capability, but we don't have a concrete need for it yet — and the Go 1.25 toolchain bump is a project-wide commitment we'd want to make on its own terms, not as a side-effect of a feature PR.

Happy to revisit if a tracing-only contribution becomes useful later (e.g., when we have a hosted-service story with SLAs). Appreciate the effort — not a reflection on code quality, which was solid.

PR #1, opened 2026-01-26
Add Telegram adapter for notifications, similar to Slack adapter. Create internal/adapters/telegram/notifier.go with SendMessage, SendTaskStarted, SendTaskCompleted, SendTaskFailed methods. Use telegram-bot-api. Add TelegramConfig to config system with bot_token and chat_id fields.

alekspetrov:
## Summary

Automated PR created by Pilot for task TASK-38051.

## Changes

Add Telegram adapter for notifications, similar to Slack adapter. Create internal/adapters/telegram/notifier.go with SendMessage, SendTaskStarted, SendTaskCompleted, SendTaskFailed methods. Use telegram-bot-api. Add TelegramConfig to config system with bot_token and chat_id fields.

PR #2408, opened 2026-04-26
fix(executor): support modern opencode message schema

myagizmaktav:
## Summary
- send modern OpenCode `/session/{id}/message` payloads with typed text parts and object-form models
- fall back to legacy string-model payloads only when server validation indicates an older schema
- add focused OpenCode backend tests for model normalization, payload generation, and retry behavior

## Verification
- `go test ./internal/executor/...`
- `go test ./...` (known baseline-unrelated failures remain in `desktop` git-graph tests on this clone)
- direct local OpenCode API check: modern payload accepted, legacy payload rejected with `path=["model"]`
- local smoke test advanced past previous immediate send failure; latest `GH-1` execution remains `running` instead of failing instantly after `OpenCode server ready`

### DIFF
--- a/internal/executor/backend_opencode.go
+++ b/internal/executor/backend_opencode.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -26,6 +27,21 @@ type OpenCodeBackend struct {
 	httpClient *http.Client
 	serverCmd  *exec.Cmd
 	serverMu   sync.Mutex
+}
+
+type openCodeModelRef struct {
+	ProviderID string `json:"providerID"`
+	ModelID    string `json:"modelID"`
+}
+
+type openCodeTextPart struct {
+	Type string `json:"type"`
+	Text string `json:"text"`
+}
+
+type openCodeMessagePayload struct {
+	Model interface{}        `json:"model,omitempty"`
+	Parts []openCodeTextPart `json:"parts"`
 }
 
 // NewOpenCodeBackend creates a new OpenCode backend.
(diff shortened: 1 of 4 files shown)


myagizmaktav:
Follow-up update: this branch now also includes a second fix for the false-negative OpenCode dependency check.

Additional change on top of the schema fix:
- health/dependency checks now resolve the active backend command from config
- for OpenCode, Pilot now inspects the configured `executor.opencode.server_command` binary instead of assuming a static version command path
- health checks prefer `opencode --version`, which matches current installs and wrapper scripts better

Verified locally after rebuilding the CLI:
- `pilot doctor` now reports `✓ opencode 1.4.6 [active backend]`
- the earlier false negative `✗ opencode not found [active backend]` is gone

Focused validation for this follow-up:
- `go test ./internal/health -run TestResolveBackendCommand_.*

myagizmaktav:
Follow-up update: this branch now also includes a second fix for the false-negative OpenCode dependency check.

Additional change on top of the schema fix:
- health/dependency checks now resolve the active backend command from config
- for OpenCode, Pilot now inspects the configured `executor.opencode.server_command` binary instead of assuming a static version command path
- health checks prefer `opencode --version`, which matches current installs and wrapper scripts better

Verified locally after rebuilding the CLI:
- `pilot doctor` now reports `✓ opencode 1.4.6 [active backend]`
- the earlier false negative `✗ opencode not found [active backend]` is gone

Focused validation for this follow-up:
- `go test ./internal/health -run 'TestResolveBackendCommand_.*|TestRunChecks_.*|TestCheckDependencies_ReturnsChecks' -count=1 -v`

alekspetrov (commented):
Did a deep verification pass on the heuristic — TL;DR: **safe to merge, with three small follow-ups suggested.** Also surfaced a separate (out-of-scope) bug in the response-parsing path tracked in #2409.

### What I verified

**1. Schema correct against OpenCode v1.4.6 source**
[`packages/opencode/src/session/prompt.ts:1717`](https://github.com/anomalyco/opencode/blob/v1.4.6/packages/opencode/src/session/prompt.ts) — `PromptInput.model` is `z.object({providerID, modelID}).optional()`, `parts` is `z.discriminatedUnion("type", [...])`. The PR's modern payload shape matches exactly.

**2. Retry is fully idempotent**
[`server/instance/session.ts:854-895`](https://github.com/anomalyco/opencode/blob/v1.4.6/packages/opencode/src/server/instance/session.ts) — Hono runs `validator("json", PromptInput)` **before** the route handler. On validation failure no `svc.prompt(...)` is invoked, no message persisted, no LLM tokens charged. Retrying the same `sessionID` after a 400 is safe.

**3. Error envelope matches what the heuristic looks for**
`hono-openapi`'s `validator()` delegates to `@hono/standard-validator`'s `sValidator`, which returns `c.json({success:false, error:[...issues], data}, 400)` — compact JSON via `c.json`, raw Zod issues passed through (the `zod` vendor case isn't sanitized). Substring `"path":["model"]` matches; lowercased message contains `expected string`.

**4. Probed against 7 realistic bodies**
Compact / pretty-printed / capitalized / nested-path collision (`["agent","model","providerID"]`) / `ProviderModelNotFoundError` / `parts[0].type` discriminator failure / Zod-3-style messages — all behave correctly. Notably, the exact-match pattern `"path":["model"]` (with the closing `]`) avoids false-positive on nested model paths.

### Suggestions (non-blocking)

**a. Gate on status code** — currently any non-200 enters the heuristic. Tighten to 400/422:
```go
if messageErr.StatusCode == http.StatusBadRequest || messageErr.StatusCode == http.StatusUnprocessableEntity { ... }
```

**b. Parse the envelope instead of substring-matching** — more durable across Zod / hono-openapi version drift:
```go
type ocIssue struct {
    Path     []json.RawMessage `json:"path"`
    Code     string            `json:"code"`
    Expected string            `json:"expected"`
    Received string            `json:"received"`
}
type ocBadReq struct{ Error []ocIssue `json:"error"` }
// retry when path == [`"model"`] && code == "invalid_type" && received == "object"
```
Substring matching works today (Hono `c.json` is compact) but fails on a future pretty-printer or any error-envelope reshape.

**c. Preserve original modern error if retry also fails** — currently the second attempt's error overwrites the first, hurting debugging when both schemas mismatch.

**d. One-line comment** in `shouldRetryOpenCodeMessageLegacy` documenting the targeted format (Zod 4 issues from hono-openapi standard-validator, `error[].path == ["model"]`).

### Out of scope (filed as #2409)
After this PR lands the rejection bug is gone, but `sendMessage` then decodes the success response into `{success, output, error}` — fields that don't exist on OpenCode v1.4.6's `{info, parts}` body. Result: `Success=true` with empty `Output`, silent downstream failure (no commit / empty PR). Tracked in #2409 — not your problem here.

Nice catch on the schema and the doctor false-negative. Whether to address (a)–(d) before merge is up to the maintainers.

PR #2411, opened 2026-04-26
feat(executor): make opencode request timeout configurable

### LINKED ISSUE #2410: Make OpenCode HTTP request timeout configurable
## Summary
Pilot currently hardcodes the OpenCode HTTP client timeout to 10 minutes in the backend. When OpenCode takes longer to return message response headers or stream setup, execution fails with:

```text
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

## Why this matters
OpenCode task execution time can vary significantly by repository size, provider latency, and task complexity. A fixed 10-minute HTTP timeout is too rigid for long-running sessions.

## Proposed behavior
Add a config option under:

```yaml
executor:
  opencode:
    request_timeout: 20m
```

Behavior:
- default remains 10m for backward compatibility
- configurable via YAML
- invalid or empty values fall back to 10m
- used by the OpenCode backend HTTP client for session/message requests

## Reproduction context
With current behavior, recent local runs failed after the schema issue was fixed, but then hit:

```text
Post \"http://127.0.0.1:44096/session/.../message\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

This indicates the payload is accepted, but the request lifetime is too short for some runs.

## Scope
This is specifically about OpenCode backend HTTP request timeout, not the broader task execution timeout routing already present under `executor.timeout`.

myagizmaktav:
## Summary
- add `executor.opencode.request_timeout` so OpenCode HTTP request lifetime is configurable
- keep backward compatibility by defaulting to 10m when unset or invalid
- cover timeout parsing and backend client wiring with focused tests

## Example
```yaml
executor:
  opencode:
    request_timeout: 20m
```

## Verification
- `go test ./internal/executor -run 'TestOpenCodeConfig|TestOpenCodeConfigEffectiveRequestTimeout|TestNewOpenCodeBackend' -count=1 -v`
- `make build`

Closes #2410

### DIFF
--- a/internal/executor/backend.go
+++ b/internal/executor/backend.go
@@ -643,6 +643,24 @@ type OpenCodeConfig struct {
 
 	// ServerCommand is the command to start the server (default: "opencode serve")
 	ServerCommand string `yaml:"server_command,omitempty"`
+
+	// RequestTimeout is the maximum time to wait for OpenCode HTTP responses.
+	// Applies to session creation, message send, and SSE response headers.
+	// Default: "10m"
+	RequestTimeout string `yaml:"request_timeout,omitempty"`
+}
+
+// EffectiveRequestTimeout returns the OpenCode HTTP request timeout.
+// Falls back to 10m when empty or invalid.
+func (c *OpenCodeConfig) EffectiveRequestTimeout() time.Duration {
+	if c == nil || c.RequestTimeout == "" {
+		return 10 * time.Minute
+	}
+	d, err := time.ParseDuration(c.RequestTimeout)
+	if err != nil || d <= 0 {
+		return 10 * time.Minute
+	}
+	return d
 }
 
 // DefaultBackendConfig returns default backend configuration.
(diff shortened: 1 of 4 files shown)


alekspetrov:
Thanks! The conflict is from commit `8663915` which duplicates the OpenCode message-schema fix that already shipped in v2.100.3 (PR #2414). Could you rebase off main and drop that commit, leaving only `8fc1c31` (the timeout config)? After rebase the diff should be limited to: `backend.go`, `backend_opencode.go` (timeout fields only), and the new `backend_opencode_test.go` cases for timeout parsing. Health-check changes can also drop if v2.100.3 already covers them.

alekspetrov:
Rebased onto current main and force-pushed. Dropped the stale schema-fix commit (`8663915`) — that was already shipped in v2.100.3 / PR #2414. Kept only the timeout config commit (`8fc1c31`). Diff is now +58/−3 (down from +442/−43). Authorship preserved.

Will need workflow approval on the Checks tab to run CI.

alekspetrov (approved):
LGTM. Adds executor.opencode.request_timeout with 10m default, validation, and tests. Rebased onto current main.

PR #2419, opened 2026-04-26
fix(telegram): cap startup API calls

### LINKED ISSUE #2417: Telegram startup can hang indefinitely on singleton/getMe calls
## Problem
`pilot start --telegram` can block during startup before polling fully comes up.

## Root Cause
Two startup API calls have no short bound:
- `CheckSingleton()`
- initial `getMe()` in `StartPolling()`

If Telegram API is slow or unreachable, startup can stall instead of degrading gracefully.

## Proposed Fix
Wrap both startup calls in a short timeout (10s) so polling startup continues or fails fast.
Add regression coverage with a slow test server.

myagizmaktav:
## Summary
- add a short timeout around Telegram singleton and initial `getMe()` startup calls
- prevent `pilot start --telegram` from hanging indefinitely when Telegram is slow or unreachable
- add a regression test with a delayed startup API response

## Verification
- `go test ./internal/adapters/telegram/...`

Closes #2417

### DIFF
--- a/internal/adapters/telegram/handler.go
+++ b/internal/adapters/telegram/handler.go
@@ -21,6 +21,8 @@ import (
 	"github.com/qf-studio/pilot/internal/transcription"
 )
 
+const startupAPITimeout = 10 * time.Second
+
 // MemberResolver resolves a Telegram user to a team member ID for RBAC (GH-634).
 // Decoupled from teams package to avoid import cycles.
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
LGTM. Surgical 10s startup timeout on CheckSingleton/getMe with regression test.

PR #2420, opened 2026-04-26
fix(executor): extend opencode self-review timeout

### LINKED ISSUE #2416: OpenCode self-review phase times out before backend completes
## Problem
OpenCode-backed runs can finish main execution but still fail during self-review because the runner cancels review too early.

## Root Cause
`runSelfReview()` uses a hardcoded 2 minute timeout even when the OpenCode backend itself allows much longer request durations.

## Impact
Tasks can appear to regress in the self-review phase despite a healthy OpenCode session.

## Proposed Fix
Use a longer OpenCode-specific self-review timeout while keeping the shorter timeout for other backends.
Add regression tests for backend-specific timeout selection.

myagizmaktav:
## Summary
- give OpenCode-backed self-review a longer timeout window than the default 2 minutes
- keep the short 2 minute self-review timeout for other backends
- add regression coverage for backend-specific self-review timeout selection

## Verification
- `go test ./internal/executor/...`

Closes #2416

### DIFF
--- a/internal/executor/runner.go
+++ b/internal/executor/runner.go
@@ -562,6 +562,17 @@ func (r *Runner) backendType() string {
 	return "claude-code"
 }
 
+// selfReviewTimeout returns the per-backend timeout for the self-review phase.
+// OpenCode runs are legitimately slower than Claude Code (server-managed
+// session, larger streaming overhead); a 2-minute cap cancels review while the
+// backend is still working and surfaces as a false regression. GH-2416.
+func (r *Runner) selfReviewTimeout() time.Duration {
+	if r.backendType() == BackendTypeOpenCode {
+		return 10 * time.Minute
+	}
+	return 2 * time.Minute
+}
+
 // fallbackModelName returns the best-known model name for telemetry rows when
 // the backend stream did not surface a model field. Used to distinguish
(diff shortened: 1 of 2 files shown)


alekspetrov:
Thanks! The diff is much larger than a self-review timeout change should need (+114/−109 in runner.go). Looks like a rebase artifact or formatter pass. Could you rebase off main and resubmit with a minimal diff covering just: (a) backend-aware self-review timeout selection, (b) the new test in runner_test.go? Happy to merge once the runner.go diff is scoped to the timeout logic.

alekspetrov:
Rebased onto current main and force-pushed (was stale by ~v2.99 work). The previous +114/−109 in runner.go was gofmt churn against an older base plus removal of post-branch work. New diff is +49/−2: `selfReviewTimeout()` helper (10m OpenCode / 2m otherwise) wired into `runSelfReview`'s context, plus 3 unit-test cases. Authorship preserved.

Will need workflow approval on the Checks tab to run CI on the new push.

alekspetrov (approved):
LGTM. Backend-aware self-review timeout (10m OpenCode / 2m otherwise) with focused tests. Rebased onto current main.

Issue #2464, opened 2026-04-30
Self-upgrade fails on Linux with ETXTBSY when replacing running binary


myagizmaktav:
## Summary
Running `pilot upgrade` on Linux can fail during install with:

```text
install failed (rolled back): failed to create binary: open /config/.local/bin/pilot: text file busy
```

## Root Cause
The upgrader writes directly to `u.binaryPath` using `os.O_TRUNC` in all install paths:
- tar.gz extraction
- zip extraction
- direct binary copy

When `pilot` upgrades itself, Linux refuses to truncate the currently executing binary and returns `ETXTBSY` (`text file busy`).

## Expected Behavior
The upgrader should write the new binary to a temp file in the destination directory and atomically rename it into place after the write completes.

## Reproduction
1. Install an older `pilot` binary in `~/.local/bin/pilot`
2. Run `pilot upgrade`
3. Confirm update is available and proceed
4. Observe install failure at the final copy step with `text file busy`

## Notes
I prepared a fix branch that routes installation through temp-file + rename and adds regression tests for atomic replacement and temp cleanup.

myagizmaktav:
Fix PR opened: #2465

- PR: https://github.com/qf-studio/pilot/pull/2465
- Branch: `fix/upgrade-atomic-install`
- Summary: switch self-upgrade installs to temp-file + atomic rename so Linux upgrades do not fail with `ETXTBSY` / `text file busy` while replacing the running binary.

PR #2465, opened 2026-04-30
fix(upgrade): replace binaries atomically during self-update

myagizmaktav:
## Summary
- write downloaded binaries to a temp file in the target directory and rename into place instead of truncating the running executable
- route tar.gz, zip, and direct-binary installs through the shared atomic installer path
- add regression tests that verify atomic replacement and temp-file cleanup on failed writes

## Why
`pilot upgrade` was failing on Linux with `open .../pilot: text file busy` because the upgrader tried to truncate the currently running executable in place. Renaming a completed temp file into place avoids `ETXTBSY` and keeps the existing binary intact if installation fails.

## Testing
- `go test ./internal/upgrade/...`


alekspetrov:
Hi @myagizmaktav — thanks for catching the ETXTBSY bug and shipping a clean atomic-replace fix. We extracted commit d2e9250 into PR #2565 with you credited as `Co-authored-by`, and it's now merged. Issue #2464 is closed.

We weren't able to take this PR as-is because it bundles 33 commits across several unrelated areas (OpenCode backend rewrite, doctor refactor, Telegram startup changes, quality gate work). Each of those looks worth reviewing on its own merits, but they need to be evaluated independently — a single PR of that size and scope is hard for us to review safely.

If you're up for it, please refile each topic as its own focused PR (≤5 commits, single concern, branched from current main). Specifically the OpenCode backend modernization and the doctor health checks both look useful. We'll review each one on its own.

Closing this PR — please don't take it as a rejection of the other work, just of the bundling. Looking forward to the resubmissions. 🙏

PR #2852, opened 2026-05-08
GH-2846: feat(autopilot): restore Prometheus metrics from SQLite snapshot on daemon start - In `internal/autopilot/metrics.go`, add `RestoreFromRow(row *memory.AutopilotMetricsRow)` that takes the mutex and copies the snapshot fields per the spec (IssuesProcessed success/failed/rate_limited, PRsMerged, PRsFailed, PRsConflicting, CircuitBreakerTrips). Skip histograms and the per-endpoint APIErrors map. In `internal/autopilot/metrics_persister.go`, call `mp.store.LatestAutopilotMetrics()` once at the top of `Run()` before the ticker loop; on success+non-nil, invoke `RestoreFromRow` and log at info; on error or nil, log at debug and continue with zero state

### LINKED ISSUE #2846: feat(autopilot): restore Prometheus metrics from SQLite snapshot on daemon st...
Parent: GH-2836

never fail-fatal. Tests: unit test `RestoreFromRow` populates every counter from a row; unit test that incrementing after restore yields `row + 1`; integration test using a real `Store` seeded with one row asserts `Snapshot()` reflects it before any tick fires; integration test with an empty store leaves counters at zero with no error. Verify with `make test`, `make lint`, and the end-to-end curl/sqlite check from the task spec.

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-2846.

Closes #2846

## Changes

never fail-fatal. Tests: unit test `RestoreFromRow` populates every counter from a row; unit test that incrementing after restore yields `row + 1`; integration test using a real `Store` seeded with one row asserts `Snapshot()` reflects it before any tick fires; integration test with an empty store leaves counters at zero with no error. Verify with `make test`, `make lint`, and the end-to-end curl/sqlite check from the task spec.

### DIFF
--- a/internal/autopilot/metrics.go
+++ b/internal/autopilot/metrics.go
@@ -3,6 +3,8 @@ package autopilot
 import (
 	"sync"
 	"time"
+
+	"github.com/qf-studio/pilot/internal/memory"
 )
 
 // Metrics collects autopilot operational metrics.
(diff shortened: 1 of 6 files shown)


alekspetrov:

Issue #3650, opened 2026-06-24
feat(comms): route operational queries to inline store-backed handler


alekspetrov:
Parent: GH-3648

In internal/comms/handler.go, add the operational check at the top of detectIntent ahead of the greeting/question fast-path; add case intent.IntentOperational to the dispatch switch; implement handleOperational to query h.store.GetActiveExecutions() + h.store.GetQueuedTasks(N), format via shared helper, and reply via SendText() with no executor.Task or runner call. Guard h.store == nil to fall back to handleQuestion. In internal/comms/commands.go, extract formatQueueSummary(running, queued []*memory.Execution) string from existing /queue and /status handlers and call it from all three sites. Add table-driven tests in handler_test.go: detectIntent ordering, handleOperational with fake store asserting inline text and runner never invoked, empty queue → queue is empty message, nil store → falls back to handleQuestion. Verify make lint and make test green.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-3648 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-3650 is already queued or running
```

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-3650 is already queued or running
```

alekspetrov:
✅ Completed as part of GH-3648
PR: https://github.com/qf-studio/pilot/pull/3654

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-3650 is already queued or running
```

alekspetrov:
✅ Pilot completed!

| Metric | Value |
|--------|-------|
| Duration | 12m52s |
| Branch | `pilot/GH-3650` |
| PR | https://github.com/qf-studio/pilot/pull/3655 |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3655 |
| Branch | `pilot/GH-3650` |
| Time to merge | 2m28s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3655 |
| Branch | `pilot/GH-3650` |
| Time to merge | 3m22s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
[...]

PR #3653, opened 2026-06-24
feat(comms,intent): add IntentOperational for live queue queries (GH-3650)

alekspetrov:
## Summary

- Adds `IntentOperational` constant + `IsOperationalQuery()` regex matcher in `internal/intent` so that phrases like "what's in the queue?", "anything running?", "how many tasks?" route to a dedicated handler rather than falling through to the executor
- Implements `Handler.handleOperational()` in `internal/comms/handler.go`: reads `GetActiveExecutions()` + `GetQueuedTasks(10)` from `h.store`, formats via the shared `formatQueueSummary` helper, and replies with `SendText()` — no `executor.Task`, no runner call
- Guards `h.store == nil` to fall back to `handleQuestion`; adds nil-runner guard in `handleQuestion` to prevent panic on that path
- Extracts `formatQueueSummary(running, queued []*memory.Execution) string` from `/queue` and `/status` handlers in `internal/comms/commands.go`; all three sites now call the shared helper

## Test plan

- [ ] `go test ./internal/comms/... ./internal/intent/...` — all pass
- [ ] `go test ./...` — all 42 packages pass
- [ ] `golangci-lint run --new-from-rev=origin/main ./...` — 0 issues
- [ ] New table-driven tests cover: detectIntent ordering (operational before greeting/question), handleOperational with seeded store (TASK-42 in reply, runner never invoked), empty queue → "Queue is empty", nil store → executor-not-available fallback

### DIFF
--- a/internal/comms/commands.go
+++ b/internal/comms/commands.go
@@ -225,11 +225,14 @@ func (c *CommandHandler) handleStatus(ctx context.Context, contextID string) {
 		}
 	}
 
-	// Queue info from memory store
+	// Queue/running info from memory store
 	if c.store != nil {
+		running, _ := c.store.GetActiveExecutions()
 		queued, err := c.store.GetQueuedTasks(10)
-		if err == nil && len(queued) > 0 {
-			sb.WriteString(fmt.Sprintf("\n📋 Queue: %d task(s)\n", len(queued)))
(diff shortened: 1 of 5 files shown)


alekspetrov:

Issue #3726, opened 2026-07-03
fix(executor,config): fail-loud pass 2 — log omitted --model routing, error on empty ${VAR} secrets at config load


alekspetrov:
Blocked by: #3725

## Context

TASK-379 (runtime self-verification) Phase A, wave V2. Two more silent degradation paths: model routing that silently omits `--model` (worker runs on whatever the user's personal Claude Code default is), and config env-var interpolation that silently turns unset `${VAR}` into empty secrets.

Parent plan: `.agent/tasks/TASK-379-runtime-self-verification.md` (Phase A: A3, A4).

Serialized after #3725 because both touch `internal/executor/runner.go` (different regions; avoiding parallel-worker rebase churn).

### 1. Model-routing silence: log when routing is skipped (A3)

- `internal/executor/backend_claudecode.go:357-361` — the `opts.Model != ""` branch logs "Using routed model"; the empty branch omits `--model` with zero log lines. Add an INFO log on the empty branch: "model routing produced no model — omitting --model; worker will use its own default" (include backend type and task ID).
- `internal/executor/runner.go:2611-2620` — the zero-telemetry `fallbackModelName()` call site has no logging (its siblings at `runner.go:2189` and `:2237` both Warn). Add `log.Warn` stating telemetry produced no model name and the recorded name is a config-derived guess.
- `internal/executor/model_routing.go:102-105` — `SelectModel` returns `""` silently when routing is nil/disabled. Add a Debug log (this fires per-task; Debug is the right level given the backend log above covers the visible case).

Behavior must NOT change: `--model` must still be omitted for the Claude Code backend when routing returns empty (GH-2450 — do not reintroduce default_model clobbering).

### 2. Fail loud on empty `${VAR}` expansion for sensitive config fields (A4)

- `internal/config/config.go:503` — `os.ExpandEnv(string(data))` silently replaces unset `${VAR}`/`$VAR` with `""`, so a daemon can boot with empty tokens and fail far from the cause.
- Before expansion, scan the raw YAML for `${VAR}` / `$VAR` references; for each variable that resolves to empty, check whether the reference appears on a line whose key matches sensitive patterns (`token`, `key`, `secret`, `password`, case-insensitive). If so, return a load error naming the variable and the config line/key. Non-sensitive empty expansions log a Warn but do not fail.
- Add a config test with an unset env var referenced by a `*_token` field asserting `Load` fails with the variable named, and one for a non-sensitive field asserting Warn-but-loads.

[...]

PR #3763, opened 2026-07-03
GH-3755: feat(config): fail loud on empty ${VAR} expansion for sensitive fields

### LINKED ISSUE #3755: feat(config): fail loud on empty ${VAR} expansion for sensitive fields
Parent: GH-3726

In internal/config/config.go around the os.ExpandEnv(string(data)) call at line 503, pre-scan the raw YAML for ${VAR} and $VAR references and check each against os.LookupEnv. For any reference that resolves to empty, inspect the YAML line it appears on: if the key matches token, key, secret, or password (case-insensitive substring match), return a load error naming both the variable and the offending config key/line; otherwise log.Warn naming the variable but proceed with expansion. Add unit tests in internal/config/: (a) github.token: ${UNSET_VAR} with UNSET_VAR unset asserts Load returns an error mentioning UNSET_VAR; (b) a non-sensitive field (e.g. a label or name) with an unset var loads successfully and emits a Warn; (c) an existing real config still loads unchanged as a regression guard. Verify with: make build && make test && make lint; go test ./internal/config/... -run 'TestLoad' -v; UNSET_TEST_VAR= go run ./cmd/pilot doctor 2>&1 | head -5

## Scope fence

Implement ONLY the slice described above. The parent issue GH-3726 is decomposed

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-3755.

Closes #3755

## Changes

GitHub Issue #3755: feat(config): fail loud on empty ${VAR} expansion for sensitive fields

Parent: GH-3726

In internal/config/config.go around the os.ExpandEnv(string(data)) call at line 503, pre-scan the raw YAML for ${VAR} and $VAR references and check each against os.LookupEnv. For any reference that resolves to empty, inspect the YAML line it appears on: if the key matches token, key, secret, or password (case-insensitive substring match), return a load error naming both the variable and the offending config key/line; otherwise log.Warn naming the variable but proceed with expansion. Add unit tests in internal/config/: (a) github.token: ${UNSET_VAR} with UNSET_VAR unset asserts Load returns an error mentioning UNSET_VAR; (b) a non-sensitive field (e.g. a label or name) with an unset var loads successfully and emits a Warn; (c) an existing real config still loads unchanged as a regression guard. Verify with: make build && make test && make lint; go test ./internal/config/... -run 'TestLoad' -v; UNSET_TEST_VAR= go run ./cmd/pilot doctor 2>&1 | head -5

## Scope fence

Implement ONLY the slice described above. The parent issue GH-3726 is decomposed

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


alekspetrov:

PR #3804, opened 2026-07-04
fix(cmd): pilot upgrade hangs forever on confirmation prompt and is unkillable by SIGTERM

### LINKED ISSUE #3791: fix(cmd): pilot upgrade hangs forever on confirmation prompt and is unkillable by SIGTERM
## Context
Defect D8 from TASK-382 (`.agent/tasks/TASK-382-restart-epic-defect-burndown.md`). Recurring: `pilot upgrade` processes hang indefinitely and survive `kill`. Observed 2026-07-03: a `pilot upgrade` started 17:58 UTC sat blocked for ~3h (0.27s CPU, state S+), survived SIGTERM, required SIGKILL — and its lingering presence disrupted two subsequent daemon restarts.

## Root cause (verified in code)
1. **Unbounded interactive prompt**: `cmd/pilot/upgrade.go:194-201` — `Proceed with upgrade? [y/N]:` blocks on `bufio.NewReader(os.Stdin).ReadString('\n')` with no timeout and no TTY check. Missed prompt / non-TTY stdin → blocks forever.
2. **Signal handler makes it unkillable**: `upgrade.go:151-157` — `signal.Notify(sigCh, SIGINT, SIGTERM)` catches the signal, prints "Upgrade cancelled", and calls `cancel()` — but the stdin read is not context-aware, so nothing unblocks. The handler downgrades SIGTERM from default-terminate to a no-op.

## Fix
- Prompt via goroutine + channel; `select` on {input, ctx.Done(), timeout (60s)} — on signal or timeout, exit with a clear message and non-zero status.
- If `os.Stdin` is not a TTY (`golang.org/x/term.IsTerminal`), do not prompt: fail with "non-interactive stdin — use --yes" (or honor `--yes`).
- After the signal handler fires and cleanup completes, the process must actually exit.
- Audit the same prompt pattern elsewhere in `cmd/pilot` (onboarding wizards) for the identical hang.

## Acceptance
- [ ] `echo | pilot upgrade` and `pilot upgrade </dev/null` exit promptly with a clear message, never hang
- [ ] SIGTERM during the prompt terminates the process within 1s
- [ ] Unanswered prompt times out (60s) with "upgrade cancelled (timeout)"
- [ ] `--yes` path unaffected
- [ ] make test && make lint pass

alekspetrov:
## Summary
Fixes GH-3791 / defect D8 from TASK-382. `pilot upgrade`'s `[y/N]` confirmation prompt blocked on a plain `bufio.Reader.ReadString` with no timeout and no TTY check, so a missed prompt or non-interactive stdin hung the process forever. The SIGTERM handler cancelled a `context.Context` that the blocking read never observed, so `kill -TERM` was a no-op — only `SIGKILL` could clear the zombie process, and its lingering presence disrupted subsequent daemon restarts.

## Changes
- `cmd/pilot/upgrade.go`: extracted the confirmation prompt into `confirmUpgrade`/`readConfirmation`. `confirmUpgrade` fails fast with a clear error if stdin isn't a TTY (honoring `--yes` instead). `readConfirmation` races the blocking read against `ctx.Done()` and a 60s timeout via `select`, so SIGTERM or an unanswered prompt now exit promptly instead of hanging.
- The SIGTERM/SIGINT handler still cancels `ctx`, but the confirmation read now actually observes it.
- Added `golang.org/x/term` (already an indirect dep via bubbletea) as a direct import for the TTY check.
- `cmd/pilot/upgrade_test.go`: unit tests for `readConfirmation` covering accepted/declined input, context cancellation, and timeout.

## Verification
- `go build ./...`, `go test ./...`, `golangci-lint run --new-from-rev=origin/main ./...` all pass.
- Manually verified against the built binary:
  - `echo | pilot upgrade` and `pilot upgrade </dev/null` exit immediately with "non-interactive stdin — use --yes" (exit 1).
  - Via a real pty: SIGTERM sent while the prompt is displayed terminates the process in ~2ms (well under the 1s requirement).
  - `--yes --force` path is unaffected and proceeds straight to download.

## Test plan
- [x] `echo | pilot upgrade` exits promptly, never hangs
- [x] `pilot upgrade </dev/null` exits promptly, never hangs
- [x] SIGTERM during the prompt terminates the process within 1s
- [x] Unanswered prompt times out (60s) with a clear cancellation message
- [x] `--yes` path unaffected
- [x] `make test && make lint` pass


alekspetrov:
## Audit: other cmd/pilot prompt sites (per issue's Fix section)

Searched for the same `bufio.Reader(os.Stdin).ReadString('\n')` pattern outside upgrade.go:

- `setup.go:481`, `project_wizard.go`, `onboard.go`, `onboard_ticket.go:763`, `init_project.go` — interactive setup/onboarding wizards
- `project.go:305` (`project remove` confirmation)
- `interactive.go` (multiple prompts: menu selection, task description, execute confirm, "press enter to continue")

NO-OP RATIONALE for these sites: unlike `pilot upgrade`, none of them run unattended or get invoked by automation/daemon/scripts — they're only reachable by a human explicitly running `pilot setup`/`pilot onboard`/`pilot interactive`/`pilot project` at an attended terminal. The GH-3791 incident (3h hang, unkillable, disrupted daemon restarts) is specific to `pilot upgrade` because it's the one command plausibly invoked from a non-interactive context (script, cron, CI) or left running with nobody watching. Applying the same TTY-check + ctx/timeout-race treatment to the wizard commands would be a much larger refactor (7 files, ~10 call sites) with no corresponding incident evidence, so it's out of scope for this fix. Flagging here in case a future defect burn-down wants to pick it up as a separate ticket.

Issue #3938, opened 2026-07-06
fix(executor): emit stage events, harvest tokens, and post honest decompositi...


alekspetrov:
Parent: GH-3924

In internal/executor/, close the fail-silent regression on the epic-parent path: Emit execution_events past spec_validated (claude_started, decomposed with child issue list or implementation_started, and a terminal event completed/failed/no_op) so pilot trace <task-id> shows the full lifecycle. Fix the token harvester so executions.tokens_output/tokens_input are populated whenever Claude is actually invoked — a 37-minute run must not persist as tokens_output=0. Add a no-op guard that classifies 'completed with zero tokens AND zero files AND no commit/PR' as failed (or explicit no_op with reason), following the GH-3224 ghost-SHA guard shape referenced in bug_pilot_ghost_closes. Replace the '⚠ no changes were made' comment on decomposed epic parents with an honest 'decomposed into N children: #a, #b, #c' summary sourced from the actual child-dispatch results. Add table-driven tests covering: (a) a fake Claude stream produces tokens_output > 0 and full event sequence; (b) zero-output run trips the no-op guard instead of marking completed; (c) decomposition emits the child-list comment; (d) existing close-guards (GH-3513 text-search confirmation, open-PR veto) remain unchanged.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-3924 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
✅ Completed as part of GH-3924
PR: https://github.com/qf-studio/pilot/pull/3940

alekspetrov:
PR #3940 was closed without merging: CI fix size guard: PR has 850 additions, over limit 200 (likely cascade contamination — escalate to human)

This issue will not be retried automatically under its own number — see the reason above for what happens next.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3947 |
| Branch | `pilot/GH-3938` |
| Time to merge | 31m28s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3947 |
| Branch | `pilot/GH-3938` |
| Time to merge | 32m21s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3947 |
[...]

Issue #3513, opened 2026-06-09
Scope TUI dashboard to a single project when -p is set (TASK-284)


alekspetrov:
## Goal

When `pilot start -p <path> --dashboard` is run, **every SQLite-backed metrics panel must show only `<path>`'s data** instead of aggregating across all projects in the store. In global mode (no `-p`), behavior is unchanged.

Today only adapter polling/execution and the git-graph panel are scoped to `-p`. The metrics cards (recent executions, lifetime tokens/cost, lifetime task counts, 7-day sparklines, in-flight task list) and the gateway HTTP API all aggregate globally.

This task was planned in Navigator (`.agent/tasks/TASK-284-dashboard-project-scope.md`) and re-verified against `main` on 2026-06-09. **The spec below is authoritative — all line anchors were freshly verified, but store/TUI line numbers drift fast, so confirm with `grep -n` before editing.**

## Mechanism (additive, low risk)

3 store methods gain an optional `projectPath string` arg. **Empty string = exact current behavior**, so every non-TUI caller passes `""` and is unaffected. The daemon's resolved `-p` path is threaded through the TUI, the gateway HTTP API, and the live in-flight task collector.

The `DashboardStore` interface change is compile-time enforced — any missed implementor fails to build before tests run.

## Activation rule

- `-p <path>` set → scoped, **always-on**, no opt-out flag.
- No `-p` (global mode) → **zero change**; all callers pass `""`, queries stay unfiltered.

---

### Step 1 — Store layer + tests
`internal/memory/store.go` — extend three methods, appending `AND project_path = ?` to the query only when `projectPath != ""`:

| Method | New signature | ~line |
|---|---|---|
| `GetRecentExecutions` | `(limit int, projectPath string)` | :714 |
| `GetLifetimeTokens` | `(projectPath string)` | :1776 |
| `GetLifetimeTaskCounts` | `(projectPath string)` | :1818 |

`internal/memory/store_test.go` — add `TestGetRecentExecutions_ProjectFilter`, `TestGetLifetimeTokens_ProjectFilter`, `TestGetLifetimeTaskCounts_ProjectFilter` with a multi-project fixture (rows for 2+ distinct `project_path` values; assert the filter returns only the requested one and that `""` returns all). Extend `TestGetLifetimeTokens_ExcludesZeroTokenRows` to confirm zero-token exclusion still applies under a filter.

### Step 2 — Non-TUI caller updates (all pass `""`, no behavior change)
[...]

Issue #3224, opened 2026-05-28
bug(executor): produces no commit when the fix contradicts 'obvious-looking' existing code — deterministic no-op on GH-3222


alekspetrov:
## Symptom

Executor runs to completion but produces **zero changes**, tripping the TASK-300 ghost-SHA guard:

```
execution failed: no new commit produced — worktree HEAD matches base branch parent
```

Deterministic — same result on all 3 retries.

## Concrete instance: GH-3222

Issue spec was maximally explicit: exact file (`internal/upgrade/restart.go`), exact line (22), exact change (`"--version"` → `"version"`), with proof the current code is wrong (`pilot --version` exits 1). Despite this, the executor made no edit, three times.

**Hypothesis:** the requested change makes code that *looks* correct. `--version` is the standard version flag for almost every CLI, so the executor's model reads `restart.go`, judges the existing line fine, and exits without editing — overriding an explicit, evidence-backed instruction because its prior on "`--version` is correct" is strong.

This is a **false-negative edit**: the model trusts its general knowledge over the issue's specific, verified claim. Fixed manually in #3223 (PR) since Pilot couldn't self-execute it.

## Why this matters

- A P0 one-liner with a perfect spec still couldn't be executed autonomously
- The ghost-SHA guard correctly prevented a ghost-close (good — TASK-300 working), but there's no recovery: the executor just retries the same no-op
- Any fix that contradicts a strong model prior is at risk of this failure mode

## Investigation starting points

- The executor prompt: does it instruct the model to treat the issue's claims as authoritative over its own judgment? Consider an explicit directive: "If the issue states current code is wrong and provides evidence, make the change even if the code looks correct — do not second-guess verified claims."
- Detection: when `no new commit produced` fires, can the executor capture the model's reasoning/transcript to confirm whether it (a) edited then reverted, (b) judged no change needed, or (c) errored silently? Right now we can't tell which.
- Recovery: 3 identical retries of a deterministic no-op is wasted compute. After the first `no new commit produced`, retrying with the same prompt won't help — consider escalating to `pilot-failed` immediately (or re-prompting with a stronger "you MUST edit" directive) rather than N identical attempts.

## Related

- Same family as the pre-execution gate brittleness in #3187 (empty PR title, rigid spec-header match) — all are autopilot/executor robustness gaps surfaced this week.
- GH-3222 (the bug that exposed this) fixed via #3223.

## Out of scope

- The actual smoke-test fix (done in #3223)
[...]

PR #3940, opened 2026-07-06
fix(executor): close epic-parent trace gap, harvest child tokens, add zero-artifact no-op guard

alekspetrov:
## Summary

Closes the fail-silent regression on the epic-parent path described in GH-3938:

- **Execution-events trace**: added `claude_started`, `decomposed`, `implementation_started`, and a generic terminal `completed` `Stage`, wired into both the epic-planning and direct-execution paths plus the dispatcher's terminal-success site (previously only fired when a PR existed). `pilot trace <task-id>` no longer dead-ends at `spec_validated` for epic parents or PR-less direct-commit runs.
- **Token harvester**: each child sub-issue's real token/cost/diff usage is now aggregated onto the epic-parent's own `ExecutionResult` instead of being discarded once its terminal status string was recorded — a long-running epic no longer persists `tokens_output=0`.
- **Zero-artifact no-op guard**: a "successful" result with zero tokens, zero files changed, and no commit/PR anywhere is reclassified as `no_op` (following the GH-3224 ghost-SHA guard shape) instead of silently completing. Scoped to epics and to direct tasks that actually expected a deliverable (`CreatePR`/`DirectCommit`), so legitimate no-code-required runs aren't misclassified.
- **Honest decomposition comment**: `ExecutionResult.ChildIssues` now carries the real dispatched/recovered child issue list, and the completion comment reports "decomposed into N children: #a, #b, #c" instead of a blank-looking table for parents that shipped no PR of their own.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — all 47 packages pass
- [x] New table-driven tests in `internal/executor/runner_gh3938_test.go`:
  - full event sequence + token persistence via a fake token-producing backend
  - `applyZeroArtifactNoOpGuard` classification table
  - `Runner.Execute` zero-output guard scoped by `CreatePR`/`DirectCommit`
  - `decomposedEventDetail` formatting
  - epic decomposition recording children + aggregating child tokens
- [x] New/updated tests in `cmd/pilot`: decomposed-children comment rendering, and a compatibility pin between the new guard's error text and the existing GH-3513 `noOpErrorMarker` text-search close-guard
- [x] Updated existing `TestDispatcher_SyntheticDispatch_SuccessEventSequence` / `TestDispatchSuccessStage` for the new unconditional terminal event

### DIFF
--- a/.agent/tasks/gh-3938.md
+++ b/.agent/tasks/gh-3938.md
@@ -1,0 +1,10 @@
+# GH-3938
+
+**Created:** 2026-07-06
+
+## Problem
+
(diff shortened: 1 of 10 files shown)


alekspetrov:
This PR was closed without merging: CI fix size guard: PR has 850 additions, over limit 200 (likely cascade contamination — escalate to human)

CI run: https://github.com/qf-studio/pilot/commit/97679512517abba94990890913a40e1a787ec6c1/checks

Issue #3939, opened 2026-07-06
fix(autopilot): auto-close epic parent when all decomposed children are merged


alekspetrov:
Parent: GH-3924

In internal/autopilot/, add a poll-cycle check that inspects epic-parent issues (those with a decomposition comment / child list), verifies every child is closed AND its PR is merged, and closes the parent with a summary comment naming the merged PRs. If any close-guard vetoes (e.g., open PR still references the parent, dependent issues remain open), log the explicit veto reason so the dedupe path can suppress re-dispatch instead of silently looping. Add table-driven tests covering: (a) all children merged → parent closes within one poll cycle; (b) one child still open → parent stays open, no veto log spam; (c) child closed but PR unmerged → veto with reason; (d) parent already closed → no-op; verify end-to-end via make build && go test ./internal/autopilot/... && make lint.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-3924 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
✅ Completed as part of GH-3924
PR: https://github.com/qf-studio/pilot/pull/3941

alekspetrov:
PR #3941 was closed without merging: CI fix size guard: PR has 592 additions, over limit 200 (likely cascade contamination — escalate to human)

This issue will not be retried automatically under its own number — see the reason above for what happens next.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3945 |
| Branch | `pilot/GH-3939` |
| Time to merge | 16m46s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3945 |
| Branch | `pilot/GH-3939` |
| Time to merge | 17m39s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #3945 |
| Branch | `pilot/GH-3939` |
| Time to merge | 19h2m57s |

PR #3941, opened 2026-07-06
feat(autopilot): close epic parents on poll cycle once children ship

alekspetrov:
## Summary
- Adds `pollCloseEpicParents`, run every 60s from `startReconciler`, which inspects decomposed epic-parent issues (native GitHub sub-issue links) and closes them once every child is BOTH closed AND has a merged PR — stricter than the existing merge-triggered `maybeCloseParentIssue` / startup-only `recoverStaleParentIssues`, which only check that no child is still open (so a child closed without shipping could previously let the parent close early).
- Closing comment names every merged PR (`closeParentWithMergedPRs`).
- Close-guard vetoes (open child, closed child with no merged PR, an open PR still referencing the parent, or a lookup failure) log an explicit reason at Warn; the routine "child still open" case logs at Debug so it doesn't spam every poll cycle while an epic is still in progress.

## Test plan
- [x] `make build`
- [x] `make lint`
- New table-driven tests in `internal/autopilot/epic_close_test.go` cover: all children merged (parent closes, PRs named in comment), one child still open (parent stays open, no WARN-level log), child closed but PR unmerged (WARN veto with reason), parent already closed (no-op), and an open PR referencing the parent (veto), plus a search-wiring test for `pollCloseEpicParents`.

GH-3939

### DIFF
--- a/internal/autopilot/controller.go
+++ b/internal/autopilot/controller.go
@@ -2887,6 +2887,11 @@ func (c *Controller) startReconciler(ctx context.Context) {
 			return
 		case <-ticker.C:
 			c.reconcileOrphanPRs(ctx)
+			// GH-3939: poll-cycle check for decomposed epic-parents whose
+			// children are all closed and merged. Runs on this 60s cadence
+			// rather than the (sometimes 10s) main CI ticker to stay well
+			// under the GitHub Search API's rate limit.
+			c.pollCloseEpicParents(ctx)
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


alekspetrov:
This PR was closed without merging: CI fix size guard: PR has 592 additions, over limit 200 (likely cascade contamination — escalate to human)

CI run: https://github.com/qf-studio/pilot/commit/08a17ce33f2bb0258621cc52892df39353600609/checks

Issue #3926, opened 2026-07-06
feat(autopilot): per-project release publish mode (workflow|api|tag_only) — publish GitHub Releases via API


alekspetrov:
## Context

Autopilot creates release tags across all configured repos (`post_merge.action: tag`), but the post-tag log line hardcodes an assumption: `"tag created (GoReleaser will create release)"` (`internal/autopilot/controller.go`, `handleReleasing`, ~:2284). Only the **pilot** repo actually has a tag-triggered release workflow. On 2026-07-06 autopilot tagged `studio-sdk v0.29.0` and no GitHub Release was ever published — studio-sdk has a CI workflow only. The SDK github client already exposes `CreateRelease` but the tag flow never uses it.

Goal: a per-project **publish mode** so autopilot either (a) leaves publishing to the repo's workflow, (b) publishes the GitHub Release itself via API, or (c) intentionally stops at the tag — and logs what it actually does.

## Design

New `ReleaseConfig.Publish` field: `"workflow"` (default when empty — preserves current behavior byte-identically), `"api"`, `"tag_only"`.

Per-project overrides use **overlay semantics** (project > env > global), unlike the env layer's full replacement: a project block sets only the fields it declares, inheriting the rest from the resolved env/global config. Rationale: a project block containing only `publish: api` must not silently disable releasing via `Enabled`'s zero value. `Trigger`, `VersionStrategy`, `RequireCI` stay env/global-only (document in the struct comment).

## Implementation

1. **`internal/autopilot/types.go`**
   - Add `Publish string` (yaml `publish`) to `ReleaseConfig` (~:351) + method `PublishMode() string` normalizing empty → `"workflow"`.
   - New `ProjectReleaseConfig` overlay struct: `Enabled *bool`, `Publish string`, `TagPrefix string`, `GenerateChangelog *bool`, `NotifyOnRelease *bool` (all `yaml:",omitempty"`), with `Apply(base *ReleaseConfig) *ReleaseConfig` — deep-copies base and overrides only set fields; `Apply(nil)` returns a config only if the overlay sets `enabled: true` (from `DefaultReleaseConfig()`), else nil.

2. **`internal/config/config.go`**
   - Add `Release *autopilot.ProjectReleaseConfig` (yaml `release,omitempty`) to `ProjectConfig` (~:232), mirroring the `Quality *quality.Config` pattern (:249). `internal/config` already imports `internal/autopilot` (:26).
[...]

PR #3949, opened 2026-07-06
feat(autopilot): overlay resolution + API-based Release publishing (GH-3926)

alekspetrov:
## Summary

Completes the per-project release publish mode (GH-3926): autopilot creates tags for all configured repos, but assumed a tag-triggered CI workflow (e.g. GoReleaser) would publish the GitHub Release. Only the pilot repo actually has one — on 2026-07-06 autopilot tagged `studio-sdk v0.29.0` and no Release was ever published, since studio-sdk has a CI-only workflow.

- `ReleaseConfig.PublishMode()` normalizes `Publish` (`workflow`/`api`/`tag_only`, default `workflow`).
- `ProjectReleaseConfig.Apply()` resolves the project > env > global overlay (a project block setting only `publish: api` inherits `Enabled`/`TagPrefix`/etc. rather than silently disabling releasing).
- `NewController` now applies options before constructing the releaser, so `WithReleaseOverride` participates in the resolved config the releaser is built from.
- `handleReleasing` dispatches its post-tag behavior on the resolved mode:
  - `workflow` (default): unchanged behavior, reworded log.
  - `api`: Pilot calls the new `Releaser.CreateReleaseForRepo` directly; 422 `already_exists` is treated as released.
  - `tag_only`: stops at the tag, skips the release-summary enrichment goroutine (which would otherwise poll for 5 min before warning).
- Idempotence fix: both release-drain paths (exact-tag match, duplicate-tag-create) now check `GetReleaseByTag` and create the release if missing, before draining the PR — so a transient `CreateRelease` failure after a successful tag never permanently loses the release.
- Removes the hardcoded `"tag created (GoReleaser will create release)"` log line.

Config validation (`internal/config`) and controller-construction wiring (`cmd/pilot/main.go`) landed in prior commits (6eea98ce, 248d80a0); this PR finishes the overlay resolution and `handleReleasing` consumption those commits were built to compile against. `configs/pilot.example.yaml` docs also landed previously (8bc3ed0f).

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/autopilot/... ./internal/config/...` — all green, including new tests: `TestProjectReleaseConfig_Apply`, `TestResolvedRelease_Precedence`, `TestHandleReleasing_PublishModes`, `TestHandleReleasing_APIModeDuplicateRelease`, `TestHandleReleasing_APIModeRetryAfterCreateReleaseFailure`
- [x] `go test ./...` (full suite) green
- [x] `grep -r "GoReleaser will create release"` — no matches


alekspetrov:
This PR was closed without merging: CI fix size guard: PR has 727 additions, over limit 200 (likely cascade contamination — escalate to human)

CI run: https://github.com/qf-studio/pilot/commit/3159d12332f080cc6687db33423804edc6934c6c/checks

PR #3955, opened 2026-07-07
feat(alerts): add release_missing alert type (GH-3952)

alekspetrov:
## Summary
- Add `AlertTypeReleaseMissing` in `internal/alerts/types.go` with a default rule (enabled, Warning severity, 30m cooldown)
- Add `EventTypeReleaseMissing`, `handleReleaseMissing` (mirroring `handleConfigError`), and the switch case in `handleEvent` in `internal/alerts/engine.go`
- Alert message and metadata carry `repo`, `tag`, `pr` so operators can trace the stalled release from the alert alone
- Fully self-contained to `internal/alerts/` — no other packages touched

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/alerts/...` (includes new `TestHandleReleaseMissing` covering fire/disabled/cooldown paths, plus updated `TestDefaultRules`)
- [x] `golangci-lint run --new-from-rev=origin/main ./internal/alerts/...` — 0 issues

Closes #3952

PR #3918, opened 2026-07-06
fix(autopilot): scope PR state per repo — add repo column migration (GH-3903-1)

alekspetrov:
## Summary

Subtask 1 of 5 for #3903 — implements the migration piece only.

- Rebuilds `autopilot_pr_state` with `repo TEXT NOT NULL DEFAULT ''` in the PRIMARY KEY (`(repo, pr_number)`), backfilling `repo` for existing rows by parsing `pr_url` (reuses `PRState.RepoOwnerAndName`).
- Rebuilds `autopilot_pr_failures` the same way, resolving `repo` via a join against the now-repo-backfilled `autopilot_pr_state` on `pr_number`; rows whose repo can't be resolved (no matching PR-state row, or an unparseable `pr_url`) are dropped — they're ephemeral retry counters.
- Fixes the `ON CONFLICT(pr_number)` targets in `SavePRState`/`SavePRFailures` to `ON CONFLICT(repo, pr_number)` to match the new composite key (both still write `repo=''` for now — scoping the Store API by repo is subtask 2).
- Migration is idempotent and safe on fresh installs (mirrors the existing `migrateAdapterProcessedPrimaryKey` pattern).

Out of scope for this PR (tracked by remaining #3903 subtasks): scoping `SavePRState`/`LoadAllPRStates`/etc. by repo, `Controller.RestoreState` filtering, the 404-eviction guard, and the two-controller collision tests.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/autopilot/...` — added `TestStateStore_MigratePRStateAndFailuresRepoColumn` (backfill + drop-unresolved) and `TestStateStore_PRStateRepoCollision` (same `pr_number` across two repos now coexists)
- [x] `go test ./...` (full suite)
- [x] `golangci-lint run --new-from-rev=origin/main ./...`

PR #3964, opened 2026-07-07
fix(stress): bound TestMemory_LargePayloads wait to stop package-timeout hangs

alekspetrov:
## Root cause

`TestMemory_LargePayloads` (stress/memory_test.go) busy-waited on `processedCount`
with **no deadline of its own**:

```go
for atomic.LoadInt64(&processedCount) < int64(numIssues) {
    time.Sleep(100 * time.Millisecond)
}
```

It relied entirely on the poller's 30s `context.WithTimeout`. But `startParallel`'s
`ctx.Done()` branch permanently stops the dispatch loop once the context expires —
so if not all 20 issues were dispatched/processed before that 30s elapsed (e.g. CI
under load, GC pauses, slow scheduling), `processedCount` never reaches its target
and this loop spins **forever**, hanging until the *package's* 10-minute test
timeout kills the whole `stress` package. Because the hang happens before the
`defer server.Close()` line is ever reached, the goroutine dump at kill-time shows
a goroutine parked in `httptest.(*Server).goServe` — exactly what was observed on
the two sniped PRs (#3918, #3955).

The three sibling memory stress tests in the same file already had this exact
problem fixed under TASK-353 via a `waitForProcessed(t, get, want, timeout)`
helper that fails the *single test* fast with `t.Fatalf` instead of spinning —
`TestMemory_LargePayloads` was just never migrated to it.

While auditing the rest of the `stress` package for the same pattern, found and
fixed an identical unbounded loop in `TestStress_GoroutineStability`
(concurrent_test.go) — same risk, same fix.

## Fix

- `stress/memory_test.go`: `TestMemory_LargePayloads` now uses the existing
  `waitForProcessed` helper (25s bound, inside the 30s poller ctx) instead of an
  unbounded loop. Added `defer cancel()` for the context (was previously only
  cancelled on the success path).
- `stress/concurrent_test.go`: same fix for `TestStress_GoroutineStability`
  (12s bound, inside the 15s poller ctx).

No test deleted or skipped; package-level 10m timeout unchanged. This makes a
stall fail the *individual* test with a clear "timed out after Xs waiting for
processed count: got N, want M" message rather than killing the whole package
and sniping unrelated PRs.

## Test plan

- [x] `go test ./stress/ -timeout 10m -count=3` — passes
- [x] `go test ./stress/ -timeout 10m -count=5 -race` — passes (135s)
- [x] `golangci-lint run --new-from-rev=origin/main ./stress/...` — 0 issues
- [x] `go build ./...` — passes

### DIFF
--- a/stress/concurrent_test.go
+++ b/stress/concurrent_test.go
@@ -433,12 +433,13 @@ func TestStress_GoroutineStability(t *testing.T) {
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
 	go poller.Start(ctx)
 
-	// Wait for all issues to process
-	for atomic.LoadInt64(&processedCount) < int64(numIssues) {
-		time.Sleep(50 * time.Millisecond)
-	}
+	// GH-3959: bounded wait instead of an unbounded busy-loop — see memory_test.go's
+	// waitForProcessed doc comment for why an unbounded loop here can hang until the
+	// package's 10m timeout kills the whole `stress` package.
+	waitForProcessed(t, func() int64 { return atomic.LoadInt64(&processedCount) }, int64(numIssues), 12*time.Second)
 
 	cancel()
 	poller.WaitForActive()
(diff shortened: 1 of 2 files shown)


alekspetrov:
This PR was closed without merging: CI checks failed (test); fix issue #3966 created to continue this work

CI run: https://github.com/qf-studio/pilot/commit/e1de379ba433994619bac0edcff5b2a729671510/checks

Issue #4033, opened 2026-07-07
fix(executor): stuck-monitor times decomposed tasks from execution start, not...


alekspetrov:
Parent: GH-4028

Fix the stuck-task monitor so stuck_for is computed from the task's execution-start timestamp (when the runner actually begins the subtask fan-out) rather than the queue-enqueue timestamp. Evidence: GH-4021 was evicted at 18:50 with stuck_for=41m0s while subtask 2 was actively running — queue time was 18:09, execution start was 18:34. Locate the timer in the stuck-monitor code path (likely internal/executor/stuck_monitor.go or the autopilot supervisor), switch its reference clock to execution.started_at, and add a table-driven test with a decomposed task whose subtasks legitimately run past the queue-time threshold but not past the start-time threshold — assert no false eviction.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-4028 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
PR #4038 was closed without merging: CI checks failed (Knowledge Graph Drift Gate); fix issue #4039 created to continue this work

This issue will not be retried automatically under its own number — see the reason above for what happens next.

alekspetrov:
✅ Completed as part of GH-4028
PR: https://github.com/qf-studio/pilot/pull/4040

alekspetrov:
PR #4040 was closed without merging: CI checks failed (Knowledge Graph Drift Gate, Knowledge Graph Drift Gate); fix issue #4042 created to continue this work

This issue will not be retried automatically under its own number — see the reason above for what happens next.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4045 |
| Branch | `pilot/GH-4033` |
| Time to merge | 6m26s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4045 |
| Branch | `pilot/GH-4033` |
| Time to merge | 7m20s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4045 |
| Branch | `pilot/GH-4033` |
| Time to merge | 50m22s |

Issue #4021, opened 2026-07-07
fix(github/sdk-poller): auto-retry path bypasses completed-execution guard — re-dispatches shipped task, orphan-row delete races waiter into false task_failed alert (GH-3992)


alekspetrov:
## Context (GH-3992, 2026-07-07, daemon v2.230.0→v2.232.1)

Retry-2 of GH-3992 completed 17:51:06 (+02:00) creating PR #4018 (merged 17:58:56 → v2.233.0, issue closed `pilot-done`). The normal dispatch path correctly skipped at 17:51:42: `Skipping re-dispatch — completed execution exists`. But at 17:56:47 — 5 minutes **after** the completed execution existed — the auto-retry path fired anyway:

```
17:51:41 INFO  Issue was processed but status labels removed, allowing retry   number=3992
17:56:47 INFO  Auto-retrying pilot-retry-ready issue   number=3992 retry_label=pilot-retry-1
17:57:04 INFO  Task queued   execution_id=fd53446c task_id=GH-3992   ← third, redundant run
18:01:46 INFO  Deleting orphan running row (task already completed)   execution_id=fd53446c
18:01:47 ERROR Failed to process issue   number=3992 error="failed waiting for execution: failed to get execution: sql: no rows in result set"
18:01:47 INFO  alert fired   rule=task_failed severity=warning delivered_to=[slack-engineering]
```

Confirmed on v2.232.1 (post-#4008/#4016), so this is distinct from the already-queued spam fix.

## Defects

1. **Auto-retry bypasses the completed-execution guard**: the `Auto-retrying pilot-retry-ready issue` path dispatches without the `HasCompletedExecution`-style check that protects normal dispatch (`internal/adapters/github/poller.go:1046,1478` equivalents on the SDK path).
2. **Retry labels not cleared on success**: `pilot-retry-1`/`pilot-retry-ready` survived the successful retry-2 + merge, arming the redundant third dispatch. (Same staleness seen on closed #4006.)
3. **Orphan-row deletion races the waiter**: `internal/executor/dispatcher.go:285` deletes the running row while the poller handler is blocked in its wait (`cmd/pilot/handler_common.go:158`); the waiter surfaces `sql: no rows` as a task failure → **false task_failed alert for shipped work**. The waiter (or deletion path) should resolve no-rows-because-completed as success/no-op.
4. Minor: the killed third run left a stale registered worktree `pilot-worktree-GH-3992-1783439824079474000`.

## Fix

[...]

Issue #405, opened 2026-02-04
feat(executor): add planning mode for epics (GH-385 Phase 2)


alekspetrov:
## Phase 2 of GH-385: Epic Roadmap Workflow

**Depends on:** Phase 1 (epic detection) ✅ merged in v0.14.3

### Goal

When an epic task is detected, run Claude Code in `--plan` mode first to generate subtasks.

### Implementation

**File:** `internal/executor/runner.go`

1. Add `PlanEpic()` method:
```go
func (r *Runner) PlanEpic(ctx context.Context, task *Task) (*EpicPlan, error) {
    // Build planning prompt
    prompt := fmt.Sprintf("Plan the implementation for this epic task. Break it down into 3-5 sequential subtasks.\n\nTask: %s\n\n%s", task.Title, task.Description)
    
    // Run Claude Code with --plan flag
    cmd := exec.CommandContext(ctx, "claude", "--plan", "-p", prompt)
    output, err := cmd.Output()
    
    // Parse output for numbered steps
    subtasks := parseSubtasks(output)
    
    return &EpicPlan{
        ParentTask: task,
        Subtasks:   subtasks,
    }, nil
}
```

2. Add parsing logic for subtask extraction:
```go
func parseSubtasks(output []byte) []PlannedSubtask {
    // Look for numbered patterns: "1. ", "Step 1:", etc.
    // Extract title and description for each
    // Return structured subtasks
}
```

**File:** `internal/executor/types.go`

3. Add new types:
```go
type EpicPlan struct {
    ParentTask   *Task
    Subtasks     []PlannedSubtask
    TotalEffort  string
    PlanOutput   string  // raw Claude output for reference
}

type PlannedSubtask struct {
    Title       string
    Description string
    Order       int
    DependsOn   []int  // indices of subtasks this depends on
}
```

### Integration Point

In `runner.go` `Execute()` method, before normal execution:
```go
if task.Complexity.IsEpic() {
    plan, err := r.PlanEpic(ctx, task)
    if err != nil {
        return nil, fmt.Errorf("epic planning failed: %w", err)
    }
    // Return plan for dispatcher to create sub-issues
    return &ExecutionResult{
        EpicPlan: plan,
        IsEpic:   true,
    }, nil
}
```

### Tests

Add tests in `runner_test.go`:
- Epic task triggers planning mode
- Plan output is parsed correctly
- Subtasks are extracted with correct order
- Empty plan returns error

### Acceptance Criteria

[...]

PR #4038, opened 2026-07-07
fix(executor): refresh epic parent's stuck-monitor entry per child execution (GH-4033)

alekspetrov:
## Summary
- `executeSubIssuesTracked` (the live GH-405/412 epic sub-issue fan-out) posted per-child progress to GitHub via `UpdateIssueProgress`, a `gh issue comment` call that never reaches the alerts engine. The parent's `taskLastProgress` entry was only touched once, right before the sequential loop started, then frozen for the entire run — each child executes under its own task ID, so its own progress never touched the parent's entry either.
- On a long epic this let `stuck_for` be measured from that one-time pre-loop timestamp instead of the most recent child's actual execution start, so the orphan-eviction sweep (4x the stuck threshold) could evict a parent whose child was still legitimately executing. Matches the GH-4021 incident: evicted with `stuck_for=41m0s` while sub-issue 2 was actively running.
- Fix: report progress under the parent's own task ID at each child's execution start (`internal/executor/epic.go`), mirroring the pattern the (unwired — `TaskDecomposer`/`decompose.go` has no `SetDecomposer` call site in production) subtask-decompose path already used correctly. This switches the stuck-monitor's effective reference clock from the queue/pre-loop timestamp to real execution activity.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/executor/... ./internal/alerts/...`
- [x] `golangci-lint run --new-from-rev=origin/main ./...`
- [x] New table-driven regression test `TestEngine_EvaluateStuckTasks_EpicParentRefreshedByChildExecution` (`internal/alerts/engine_test.go`) reproduces the exact GH-4021 timing (`stuck_for=41m0s`) and asserts no false eviction when a child is actively executing past the queue-time threshold but not the execution-start threshold.
- [x] New wiring test `TestExecuteSubIssuesTracked_TouchesParentProgressPerChild` (`internal/executor/epic_test.go`) confirms the parent's task ID receives a progress event per child start.

### DIFF
--- a/.agent/knowledge/memories/pitfalls/mem-089.md
+++ b/.agent/knowledge/memories/pitfalls/mem-089.md
@@ -1,0 +1,24 @@
+# Pitfall: Epic parent's stuck-monitor entry freezes once sub-issue execution starts — false eviction on long sequential epics
+
+## Summary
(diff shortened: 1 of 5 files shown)


alekspetrov:
This PR was closed without merging: CI checks failed (Knowledge Graph Drift Gate); fix issue #4039 created to continue this work

CI run: https://github.com/qf-studio/pilot/commit/dbb350a146b46f7de48608dc5fd505b795672a39/checks

PR #4316, opened 2026-07-14
fix(executor): round up detectSameComponentFromTitles threshold for n=2 epics (GH-4304)

alekspetrov:
## Summary

Fixes the real root cause of the recurring `[canary:epic-lifecycle] pipeline stall` (issue #4265): `detectSameComponentFromTitles` in `internal/executor/epic.go` computed its ">=80% of titles" threshold as `int(len(subtasks) * 0.8)`, which truncates to `1` when `len(subtasks) == 2`. Any word appearing in even ONE of two titles then satisfies `count >= 1`, so `isSinglePackageScope` returned `true` for essentially every 2-child epic plan — decomposition was silently skipped (`decomposition_skipped reason=single_package_scope`) and the epic direct-executed instead.

The epic-lifecycle canary always plans exactly 2 children (version bump + changelog entry, touching unrelated root-level files), so it has hit this collapse on every scheduled run since it started — independent of whatever PR happened to merge just before each tick. GH-4304 was auto-filed against PR #4299 (a dashboard-only HISTORY-ladder change with no relation to epic planning) purely because it was the most recent merge before the failing scheduled run; verified via the ledger (`decomposition_skipped` event on execution `GH-45`) and confirmed with a standalone repro that `detectSameComponentFromTitles` returns `true` for two titles ("Version bump…" / "Changelog entry…") that share no significant word.

- Round the threshold up via integer-math ceiling (`(n*4+4)/5`) instead of truncating, so ">=80%" is enforced correctly at every subtask count, including n=2.
- Added regression test cases for exactly 2 subtasks: unrelated titles (must NOT trigger single-scope) and titles sharing a component word (must still trigger).
- Filed a pitfall memory (`.agent/knowledge/memories/pitfalls/bug_two_subtask_epic_never_decomposes.md`) documenting the `int()` truncation trap and the test-coverage gap (no existing fixture used n=2).

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/executor/...` (full package, including new `TestDetectSameComponentFromTitles` cases)
- [ ] Next scheduled `epic-lifecycle` canary run should show >=2 children created and assertions passing, once the daemon runs a build containing this commit

### DIFF
--- a/.agent/knowledge/memories/pitfalls/bug_two_subtask_epic_never_decomposes.md
+++ b/.agent/knowledge/memories/pitfalls/bug_two_subtask_epic_never_decomposes.md
@@ -1,0 +1,16 @@
+---
+name: Two-subtask epics never decompose — int(n*0.8) threshold collapse (GH-4304)
+description: detectSameComponentFromTitles in internal/executor/epic.go truncated its ">80% of titles" threshold with int(), which collapses to 1 for n=2 subtasks — any word in even ONE of two titles then satisfied count>=1, so every 2-child epic was misclassified single-package-scope and direct-executed instead of decomposed. Fixed by rounding the threshold up (ceil via integer math). Recurrence risk if any similar ">=X% of N" gate reintroduces int() truncation for small N.
+type: pitfall
+---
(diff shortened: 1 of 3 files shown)


alekspetrov:
This PR was closed without merging: CI checks failed (Knowledge Graph Drift Gate); fix issue #4317 created to continue this work

CI run: https://github.com/qf-studio/pilot/commit/57e849d40ea002d9f887fb0c1708e1b95cac1421/checks

Issue #4332, opened 2026-07-15
fix(briefs): deflake 'sql: database is closed' in internal/briefs tests — red main killed the 07-15 release train


alekspetrov:
## Context

`internal/briefs` test flaked on main (run 29378523581, commit 02896358, 2026-07-15T00:20Z): `FAIL ... sql: database is closed`. This single red `test` check burned all 5 scope-release attempts within 7 minutes and killed the 07-15 daily release train (see #4331 for the recovery-path fix — this ticket is the trigger, that one is the disease).

## Scope

1. Reproduce: `go test -race -count=20 ./internal/briefs/...` — identify the teardown/lifetime bug (classic shape: shared `*sql.DB` closed by an earlier test's cleanup while a background goroutine or subtest still uses it; or `t.Cleanup` ordering vs `defer db.Close()`).
2. Fix the root cause — per-test DB handles or properly sequenced cleanup; no `time.Sleep` band-aids, no test skips.
3. If the flake spans other packages sharing the same test-DB helper, fix the helper.

## Gate commands

```bash
go test -race -count=20 ./internal/briefs/...
make test
make lint
```

## Acceptance

- 20 consecutive `-race` runs green locally and in CI.
- No reduction in assertion strength.

alekspetrov:
PR #4343 was closed without merging: CI checks failed (Knowledge Graph Drift Gate); fix issue #4344 created to continue this work

This issue will not be retried automatically under its own number — see the reason above for what happens next.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4345 |
| Branch | `pilot/GH-4332` |
| Time to merge | 12m30s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4345 |
| Branch | `pilot/GH-4332` |
| Time to merge | 13m22s |

alekspetrov:
held for scope release schedule

PR #4343, opened 2026-07-15
fix(memory): persist exec.CreatedAt and fix sqlite time-bind format (GH-4332)

alekspetrov:
## Summary

- Fixes the actual bug behind the internal/briefs CI flake cited in GH-4332 (run 29378523581). The issue text quotes `sql: database is closed`, but pulling the real CI log shows that string belongs to an unrelated package's output block in the same `go test ./...` run — `internal/briefs`'s actual failure in that run was `TestGeneratorGenerate: generator_test.go:100: expected 1 blocked task, got 0`.
- Root cause: `SaveExecution`'s INSERT never listed `created_at`, so SQLite's `DEFAULT CURRENT_TIMESTAMP` silently overrode any test-supplied `Execution.CreatedAt`. The period queries (`GetExecutionsInPeriod`, `GetBriefMetrics`) filter `created_at` against a Go-captured `time.Time` window — under CI load, real insertion time could cross a wall-clock second after the test captured its window, dropping the row. Fixed by binding `exec.CreatedAt` (falling back to `time.Now()` when unset, matching prior behavior for production callers that never set it).
- That fix surfaced a second, deeper bug: `modernc.org/sqlite` writes a bound `time.Time` param using Go's `time.Time.String()` format by default, which SQLite's `date()`/`datetime()` can't parse (returns NULL) and which sorts inconsistently against `CURRENT_TIMESTAMP`-populated rows in plain string comparison. Fixed by opening the DB with `_time_format=sqlite` in the DSN, switching the driver to one of SQLite's own recognized date formats for every future bound `time.Time`.
- Added a Navigator pitfall memory (`.agent/knowledge/memories/pitfalls/pitfall_sqlite_time_bind_default_format.md`) documenting both the CI-log-misattribution lesson and the driver gotcha.

## Test plan

- [x] `go test -race -count=20 ./internal/briefs/...` — green (135s)
- [x] `go test -race ./internal/memory/...` — green (previously red on `TestGetDailyMetrics_ExcludesZeroTokenRows`/`_ExcludesCanaryRows` with only the created_at fix; green after adding `_time_format=sqlite`)
- [x] `make test` — green
- [x] `golangci-lint run --new-from-rev=origin/main ./...` — 0 new issues (plain `make lint` surfaces one pre-existing unrelated `unused` finding in `internal/dashboard/grom_chrome.go`, unchanged by this PR)

### DIFF
--- a/.agent/knowledge/memories/pitfalls/pitfall_sqlite_time_bind_default_format.md
+++ b/.agent/knowledge/memories/pitfalls/pitfall_sqlite_time_bind_default_format.md
@@ -1,0 +1,22 @@
+# Pitfall: modernc.org/sqlite silently mis-formats bound time.Time params, breaking date()/created_at filters
+
+## Summary
+Binding a Go `time.Time` value directly as a `database/sql` query parameter against `modernc.org/sqlite` writes it using Go's `time.Time.String()` format (`"2006-01-02 15:04:05.999999999 -0700 MST"`) by default — a format SQLite's own `date()`/`datetime()`/`strftime()` don't recognize (they return NULL) and that sorts inconsistently against `DEFAULT CURRENT_TIMESTAMP`-populated columns in plain `>=`/`<` string comparisons. Fix: open the DB with `?_time_format=sqlite` in the DSN so the driver writes one of SQLite's own recognized date formats instead.
+
+## Context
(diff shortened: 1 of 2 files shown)


alekspetrov:
This PR was closed without merging: CI checks failed (Knowledge Graph Drift Gate); fix issue #4344 created to continue this work

CI run: https://github.com/qf-studio/pilot/commit/12ed47447b09b1ffd505d4222dc70f1a4ff296ae/checks

Issue #4387, opened 2026-07-16
Guard executor worktrees against deleting graph-indexed memory files


alekspetrov:
# Guard executor worktrees against deleting graph-indexed memory files

**Created**: 2026-07-16 · **Status**: 📝 READY TO DISPATCH

## Problem

Executor sessions repeatedly delete `.agent/knowledge/memories/**` files that ARE
indexed in `.agent/knowledge/graph.json`, judging them "unindexed" during
end-of-run hygiene. The graph node survives, the file is gone → the Knowledge
Graph Drift Gate (`scripts/check-graph.py`, "Broken file links") fails the PR's
CI → autopilot approval never fires → the issue wedges in awaiting.

There is **no in-repo code** that produces these deletions — it is the Claude
session's own judgment at finalize time (commit messages like
`chore(memory): strip unindexed memory doc(s) added during execution`).
Nothing deterministic stops it.

## Evidence (recurring class, 3+ incidents)

- **PR #4385** (GH-4376): commit `28c54ef4` deleted `mem-158.md` (indexed on
  main since #4373) → drift gate red, issue stuck awaiting. Fixed manually by
  revert `8db96c2b` 2026-07-16.
- **PR #4375**: had to restore `pitfall_sqlite_time_bind_default_format.md`
  (mem-153) + `pitfall_noop_terminal_state_invisible_to_dispatch_guards.md`
  (mem-154) — indexed in graph.json but stripped/never committed by earlier
  executions (#4345 / #4350 era).
- Drift-gate history: bd750650 / #4355 ("drift gate red on main since #4350").

## Required behavior

Add a **deterministic, in-tree guard** in the executor's finalize path
(before push / PR creation) that:

1. Detects, in the worktree's staged/committed diff vs the branch base, any
   **deletion** of a file under `.agent/knowledge/memories/` that is still
   referenced by a node in `.agent/knowledge/graph.json` (the file/path field
   of `nodes.memories.*`).
2. On detection: **restore the file(s)** (fail-safe — `git checkout <base> -- <path>`
   or equivalent) and log a WARN naming the file(s) and the graph node id(s);
   alternatively hard-fail the push with an actionable error. Restoring is
   preferred: it converts a wedged-PR class into a no-op.
3. Emits an execution event (existing `recordExecutionEvent` machinery) so the
   intervention is visible in `pilot trace` / the ledger.

### Suggested placement

- `internal/executor/git.go` push/PR-create path (same layer as the pre-push
[...]

PR #4419, opened 2026-07-17
feat(executor): guard finalize path against deleting graph-indexed memory files (GH-4387)

alekspetrov:
## Summary

Executor sessions have repeatedly deleted `.agent/knowledge/memories/**` files that are still indexed in `.agent/knowledge/graph.json` during end-of-run hygiene (commit messages like `chore(memory): strip unindexed memory doc(s) added during execution`), leaving a dangling graph node that fails the Knowledge Graph Drift Gate's "Broken file links" check and wedges the PR in awaiting-approval (PR #4385, #4375).

- Adds `GitOperations.RestoreDeletedIndexedMemoryDocs` (`internal/executor/git.go`): diffs the branch against base for deletions under `.agent/knowledge/memories/`, resolves the indexed set from `graph.json` **as of base** (handling both the current `file` key and the legacy `path` key), restores any deleted-but-indexed file via `git checkout <base> -- <path>`, and commits the restoration.
- Wired into all three finalize/push paths — direct (`runner.go`), epic (`runner.go`), and decomposed-parent (`runner_decompose.go`) — logging a WARN naming the restored file(s)/node id(s) and recording a new `memory.StageMemoryGuardRestore` execution event so the intervention is visible in `pilot trace`/the ledger.
- No-ops when `graph.json` is absent at base, or when nothing indexed was deleted.

## Test plan

- [x] Unit test: indexed memory doc deleted (`file` key) → restored, no net diff vs base
- [x] Unit test: genuinely unindexed memory doc deletion is left alone
- [x] Unit test: legacy `path`-key nodes are also protected
- [x] Unit test: no-op when `graph.json` is absent at base
- [x] Unit test: no-op when nothing under `memories/` was deleted
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./...` green

## Note for reviewer

`scripts/check-graph.py` (the CI drift gate) is currently failing on `main` (`origin/main` @ `2478a1f4`) independent of this PR — 3 memory docs were stripped by an unrelated commit (`6ca5c9c4`, PR #4410) without removing their `graph.json` nodes, and 2 edges use a legacy `source`/`target`/`relation` schema the checker doesn't recognize as `from`/`to`/`type`. Both pre-date this branch and are out of scope here (the guard added in this PR prevents *future* occurrences of the first class, but doesn't retroactively repair already-merged drift or the unrelated edge-schema issue) — flagging for a separate follow-up.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/.agent/system/FEATURE-MATRIX.md
+++ b/.agent/system/FEATURE-MATRIX.md
@@ -1,6 +1,6 @@
 # Pilot Feature Matrix
 
-**Last Updated:** 2026-07-17 (v2.151.0)
+**Last Updated:** 2026-07-16 (v2.151.0)
 
 ## Legend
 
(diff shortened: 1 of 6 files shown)


alekspetrov:
This PR was closed without merging: CI fix size guard: PR has 243 production additions, over limit 200; 2 bookkeeping additions excluded, 184 test additions excluded (likely cascade contamination — escalate to human)

CI run: https://github.com/qf-studio/pilot/commit/0fc4fd6e45500146cf4e0be062fbc380ba27e57a/checks

Issue #4437, opened 2026-07-17
refactor(autopilot): extract shared check-runs current-state polling helper


alekspetrov:
Parent: GH-4415

Refactor the check-runs 'current state' read logic out of the fresh-PR registration path into a shared helper function within internal/autopilot/ so it can be reused by both the fresh registration path and the restore-on-boot path.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-4415 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
✅ Completed as part of GH-4415
PR: https://github.com/qf-studio/pilot/pull/4445

alekspetrov:
PR #4445 was closed without merging: CI checks failed (test); fix issue #4446 created to continue this work

This issue will not be retried automatically under its own number — see the reason above for what happens next.

Issue #4436, opened 2026-07-17
fix(autopilot): confirm restore-path bug in check-runs polling


alekspetrov:
Parent: GH-4415

Inspect internal/autopilot/controller.go RestoreState and ci_monitor.go arm logic to confirm hypothesis: watchers re-armed via RestoreState only listen for new check-run events and skip the initial current-state poll that fresh registrations perform (per the #4408 fix), causing already-completed check-runs to go undetected until the 30m timeout.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-4415 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
✅ Completed as part of GH-4415
PR: https://github.com/qf-studio/pilot/pull/4443

alekspetrov:
PR #4443 was closed without merging: CI checks failed (test); fix issue #4444 created to continue this work

This issue will not be retried automatically under its own number — see the reason above for what happens next.

Issue #4415, opened 2026-07-17
fix(autopilot): restored waiting_ci PRs never poll existing check-runs — already-green PRs re-fail at 30m timeout after every restart (canary 87/90/94, v2.241.2)


alekspetrov:
## Symptom

On v2.241.2 (which carries the #4408 check-runs fix, merged via #4384's PR), canary-sandbox PRs 87/90/94 timed out again at 14:53:55Z:

```
WARN msg="CI timeout" component=autopilot pr=94 waited=30m4s
INFO msg="PR stage transition" pr=94 from=waiting_ci to=failed env=stage
```

But the checks exist and PASS — `gh pr checks 87` → `test  pass  24s`. Meanwhile pointer PRs #29/#31 progressed through waiting_ci→released under the same binary, so the check-runs fix works for pointer but not canary.

## Hypotheses to check

1. The re-armed waiting_ci state (RestoreState after the 14:22Z restart) may take a different code path than fresh PR registration — the timeout fired exactly ~30m after boot, suggesting the restored watcher never actually polled check-runs before its timer expired.
2. The canary check-run completed long before the watcher (re)started; if the watcher only reacts to *new* check-run events rather than reading current state on arm, an already-green PR never produces an event and dies at timeout.
3. Repo-specific: check-run name/filter mismatch for the canary workflow.

Hypothesis 2 fits both facts (canary checks pre-completed; pointer checks ran fresh during the watch window).

## Impact

Every already-green PR that enters/re-enters waiting_ci after a restart fails at 30m. This has burned canary PRs 87/90/94/98 at least 3 times today, and blocks the epic-lifecycle canary loop (proof chain for #4265/TASK-407).

## Refs

- #4384 (original diagnosis), #4408 (the fix that covers fresh PRs), box v2.241.2, 14:53Z evidence.

alekspetrov:
🚀 decomposed into 7 children: #4436, #4437, #4438, #4439, #4440, #4441, #4442 — starting sequential execution

alekspetrov:
⏳ Progress: 0/7 - Starting: **fix(autopilot): confirm restore-path bug in check-runs polling** (4436)

alekspetrov:
⏳ Progress: 1/7 - Starting: **refactor(autopilot): extract shared check-runs current-state polling helper** (4437)

alekspetrov:
[...]

PR #4445, opened 2026-07-18
refactor(autopilot): extract shared check-runs current-state read helper (GH-4437)

alekspetrov:
## Summary

- `internal/autopilot/ci_monitor.go` independently inlined `m.ghClient.ListCheckRuns(ctx, m.owner, m.repo, sha)` in five separate methods (`checkStatus`, `GetFailedChecks`, `GetFailedCheckLogs`, `GetCheckLogs`, `GetCheckStatus`).
- Extracted `CIMonitor.fetchCheckRuns(ctx, sha)` as the single shared check-runs current-state read, and switched all five call sites to use it.
- `checkStatus` is the function both a freshly-registered PR (`OnPRCreated`) and a restored-on-boot PR (`RestoreState`) converge on via `ProcessPR` → `handleWaitingCI` → `CheckCI`, so the new shared helper is exercised identically by both paths at the point where check-runs are actually fetched from GitHub.

Investigation note: neither `OnPRCreated` nor `RestoreState` reads check-runs directly today — both only populate `c.activePRs`, walked uniformly by the shared `ProcessPR` loop — so there was no per-registration-path duplication to remove at that layer (consistent with sibling GH-4436's finding for the same epic, GH-4415). The real, literal duplication of the check-runs "current state" read was inside `CIMonitor` itself, which this PR extracts into a shared helper as requested.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l internal/autopilot/ci_monitor.go` (clean)

### DIFF
--- a/.agent/tasks/gh-4437.md
+++ b/.agent/tasks/gh-4437.md
@@ -1,0 +1,84 @@
+# GH-4437
+
+**Created:** 2026-07-17
+
+## Problem
+
+Refactor the check-runs 'current state' read logic out of the fresh-PR registration path into a shared helper function within internal/autopilot/ so it can be reused by both the fresh registration path and the restore-on-boot path.
+
+## Investigation Result: premise does not hold — NO-OP
+
+This subtask (child 2/7 of epic GH-4415) assumes the fresh-PR registration
+path (`OnPRCreated`) contains its own check-runs "current state" read logic
+that is duplicated/unavailable to the restore-on-boot path (`RestoreState`).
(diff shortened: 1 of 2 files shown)


alekspetrov:
This PR was closed without merging: CI checks failed (test); fix issue #4446 created to continue this work

CI run: https://github.com/qf-studio/pilot/commit/54a314aaa3b9f57781ab00d92187cfe5331b8e91/checks

Issue #4784, opened 2026-08-07
feat(gateway): enforce bearer auth on /api/v1 in production when a gateway token is configured (PR#4752 review)


alekspetrov:
## Problem

The gateway's bearer auth exists (`Authenticator.Middleware`, `internal/gateway/auth.go:102`) and PR#4752's composed test asserts 401s — but production never enables it: routes are wrapped only when `s.auth != nil` (`internal/gateway/server.go:244-249`), and the only non-test callers of `NewServerWithAuth` are `NewServer` itself (passing nil) and tests. Both production construction sites — `internal/pilot/pilot.go:441` (gateway mode) and `cmd/pilot/main.go:2242` (polling mode) — use `NewServer`, so the deployed daemon serves `/api/v1/` with no auth middleware at all. PR#4752 put the first state-mutating endpoint (approving a pre-merge gate) on that surface.

Today's mitigant is the default `127.0.0.1` bind (`internal/config/config.go:445`). But hosted instances must bind reachable-from-console, and the console proxy ALREADY sends `Authorization: Bearer <PILOT_GATEWAY_TOKEN>` (pilot-console `internal/proxy/proxy.go:336`, per-org SSM SecureString) that the daemon currently ignores. Any network peer that can reach a non-loopback gateway can approve PR merges.

## Acceptance

1. A config field enables auth: verify what the existing `Authenticator` binds to first; if nothing, add e.g. `gateway.auth_token` with an env override. When non-empty, BOTH production construction sites (gateway mode + polling mode) build via `NewServerWithAuth` — the GH-4738 lesson: wire both, prove it with a composed test.
2. Empty token → current behavior exactly (no auth, loopback default) — fully backwards compatible.
3. Hosted-path check: confirm the tenant config render (pilot-console `internal/fleet/configrender.go`) can carry the same `PILOT_GATEWAY_TOKEN` the console proxy sends; note in the PR description if a console-side follow-up is required (do NOT change the console repo here).
4. Composed wiring test: token configured → 401 without bearer on a read route AND the decision route; correct bearer → 200.
5. `make build` / `make test` / `make lint` green.

## Scope fence

No new auth mechanism (bearer compare only) · no console-repo changes · no bind-address changes.

**This task must NOT be decomposed — implement as a single PR.** 

---

[...]

Issue #4778, opened 2026-08-06
fix(github): remaining in-tree static-token client sites + webhook-mode startup token validation (PR#4764 review)


alekspetrov:
## Problem

Post-merge review of PR#4764 (GH-4755) confirmed the webhook-mode fix is correct, but found residual same-class defects in tree:

1. **`cmd/pilot/main.go:2863`** — static `github.NewClient(token)` in `runPollingMode` is held past construction by `NewMergeWaiter` (`:3079`, daemon-lifetime merge-wait callback) and `NewCleaner` (`:3122`, periodic stale-label loop). Violates `newGitHubClient`'s own doc contract; frozen token 401s after App-token rotation. One-line swap to `newGitHubClient(cfg)`.
2. **Test doesn't pin the load-bearing property** — `internal/pilot/github_client_test.go:75-90` asserts `p.githubClient == client` but never asserts `githubWH`/`githubNotify` (the actual consumers, `pilot.go:378-385`) captured the injected client. If option application moves back after adapter init, the regression returns and this test stays green. Also `gotAuth` (`:64-66`) is appended in handler goroutines without sync.
3. **Webhook/gateway mode never validates the token at startup** — `validateGitHubToken` (`:2864`) and the GH-3769 preflight (`:2845-2846`) run only in `runPollingMode`. A dead App key in webhook mode fails silently at first use instead of loudly at boot.

NOT in scope here (SDK-gated, tracked separately): `poller_github.go:228,478` and `main.go:2328-2331` hold static tokens in studio-sdk clients — the SDK has no TokenFunc API yet (sdk issue filed; pilot wiring leg follows the SDK release).

## Acceptance

1. Merge-waiter/cleaner client swapped to `newGitHubClient(cfg)`; no in-tree `github.NewClient(` construction whose result outlives the call remains (grep-assert in a test if cheap).
2. Rotation test asserts the consumers (`githubWH`/`githubNotify`) hold the injected client; `gotAuth` access synchronized.
3. Webhook/gateway mode runs startup token validation with the same loud-401 behavior as polling mode.
4. `make build/test/lint` green.

**This task must NOT be decomposed — implement as a single PR.**

alekspetrov:
🤖 **Pilot started working on this issue**

Task ID: `GH-4778`

I'll post updates as I make progress.

alekspetrov:
🧹 **Pilot cleanup**: Removed stale `pilot-in-progress` label.

[...]

PR #4793, opened 2026-08-07
fix(gateway): enforce bearer auth on /api/v1 in production when configured

alekspetrov:
## Summary

PR#4752 added `Authenticator.Middleware` and a gateway-package test proving it 401s, but neither production construction site ever called `gateway.NewServerWithAuth` with a real token — both called `gateway.NewServer`, which always passes `nil`. The deployed daemon served all of `/api/v1` (including the pre-merge approval decision route) with zero auth, mitigated only by the default loopback bind.

- Adds `Config.GatewayAuthConfig()` (`internal/config/config.go`) as the single gate: returns a non-nil `*gateway.AuthConfig` only when `auth.type` is explicitly `api-token` with a non-empty token. `DefaultConfig`'s `claude-code` seed and an empty token both keep today's fully-open behavior — fully backwards compatible.
- Wires **both** production construction sites through this gate:
  - `internal/pilot/pilot.go` (gateway mode)
  - `cmd/pilot/main.go` (polling mode)
- `auth.token` supports the existing `${ENV_VAR}` expansion, so a hosted daemon can set `token: "${PILOT_GATEWAY_TOKEN}"` to match the bearer token pilot-console's proxy already sends.
- Documents the new field in `configs/pilot.example.yaml`.
- Adds a composed test (`internal/config/gateway_auth_test.go`) that builds a `*config.Config` (not a hand-built `AuthConfig`) and proves 401-without-bearer / 200-with-bearer through a real HTTP round trip on both a read route and the decision route, plus a unit test for the gate logic and a test proving the empty-token path stays fully open.

## Console follow-up

Hosted daemons need `auth.token: "${PILOT_GATEWAY_TOKEN}"` set in their rendered config for this to take effect — confirming/adding that to pilot-console's tenant config render (`internal/fleet/configrender.go`) is a **console-repo follow-up**, not done here (out of scope per the issue's scope fence).

## Scope note

This PR was previously flagged by the intent judge for apparent scope creep (GH-4778 files in the diff manifest). That was a false positive: the manifest was generated by diffing against a stale local `main` ref that predated PR#4789 (which merged GH-4778's files to `origin/main`). Diffing against the true merge-base / `origin/main` shows this PR touches exactly the 8 files relevant to GH-4784, with no GH-4778 content. The branch has since been rebased onto current `origin/main`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/config/... ./internal/pilot/... ./internal/gateway/... ./cmd/pilot/...` — all pass
- [x] `go test ./...` — full suite green
- [x] `golangci-lint run` on touched packages — 0 issues
- [x] Composed wiring test proves 401 without bearer / 200 with correct bearer on both a read route and the decision route
- [x] Empty-token test proves backwards compatibility (stays fully open)


alekspetrov:
## Pre-merge review (Navigator, 2026-08-07) — APPROVE recommended

Reviewed against #4784. All 5 acceptance items met:

1. **Wiring**: reused the existing (dead) `cfg.Auth` field rather than adding a new one — `Config.GatewayAuthConfig()` is a single shared gate, and BOTH production sites (`internal/pilot/pilot.go` gateway mode, `cmd/pilot/main.go` polling mode) now build via `NewServerWithAuth`. Correct read of the spec's "verify what the existing Authenticator binds to first".
2. **Backwards compat**: nil unless explicit `api-token` + non-empty token; the `claude-code` default seed stays open. Proven by `TestGatewayAuthConfig_EmptyTokenStaysOpen`.
3. **Fail-closed on missing env var**: `token: "${PILOT_GATEWAY_TOKEN}"` with the var unset either fails `Validate()` (empty after expansion) or enforces an unguessable literal — never silently open. Good shape.
4. **Composed test starts from `*config.Config`** (not a hand-built `AuthConfig`), 401/200 on both a read route and the decision route with a real store + approval manager — exactly the lesson the accompanying `unwired-config-field-validated-but-dead` pitfall memory encodes.
5. Console follow-up correctly noted, not done here.

**Why this PR is parked**: the scope-drift gate escalated on PR title type `fix` vs issue title type `feat` — a conventional-commit prefix quibble, not actual drift (same subsystem, same work as specced). Approve via the pending Telegram ask or `gh pr merge 4793 --squash`.

One forward-looking note (no change requested): `gateway.NewServer` still exists and hardcodes `nil` auth — a future production call site can silently reintroduce this bug. Candidate for the TASK-459 Phase 4 grep-gate family (ban `gateway.NewServer(` outside tests) or deprecate it in favor of the WithAuth variant.

Issue #4791, opened 2026-08-07
feat(autopilot): platform-outage breaker part 1 — cross-PR correlation + suppress destructive actions


alekspetrov:
## Problem

During the 2026-08-06 GitHub Actions outage the daemon acted on false CI signals for ~50 minutes — closing a correct PR, spawning junk fix issues, burning retries — until a human stopped it. Per-PR guards cannot catch this: the existing circuit breaker is deliberately per-PR ("so one bad PR doesn't block others", `controller.go:108-109`) and each PR looked individually normal. The missing signal is **correlation**: several unrelated PRs failing within minutes is a platform event.

This is **part 1 of 2** (split from #4780, which timed out at 1h as a single unit — keep this leg small). Part 2 adds the external status probe, admission pause, and held-PR re-drive.

## Scope — part 1 only

1. **Tracker** following the `internal/ghbudget/ghbudget.go` shape (process-wide state, `Allow`-style gate, exactly-one log per state transition): record each CI-failure observation as `(pr, repo, class, time)`. Open the breaker when ≥N distinct PRs (default 3) observe infra-or-unknown-class failures inside a window (default 15 min). Close it after a quiet period (default 20 min with no new infra-class failure) — simple time-based recovery only.
2. **Consume the classes shipped in #4787** (`FailureClassInfra` / `FailureClassUnknown`, `internal/autopilot/failure_class.go`).
3. **While open, suppress irreversible actions in the CI-failure path** (`handleCIFailed`, `internal/autopilot/controller.go`): no `ClosePullRequest`, no fix-issue creation, no `escalateAndHold`. The PR stays at its current stage and is re-examined on a later tick — polling, board sync, gauges and running executions all continue untouched.
4. **Alert once on open and once on close** with the correlated PR list, following an existing once-per-transition pattern (`alertBillingRefusalOnce`, `controller.go:1216`, or the ghbudget transition-log shape). Never one alert per affected PR.
5. Metrics: breaker-state gauge + trip counter (mirror `RecordCircuitBreakerTrip`).
6. Config on `autopilot.Config` (`internal/autopilot/types.go:142-165` neighbourhood): enable flag + the two thresholds + quiet period; documented in `configs/pilot.example.yaml`. Static at boot is fine (`Config.Reload` has zero callers).
[...]

Issue #4683, opened 2026-08-03
fix(autopilot): self-upgrade drain deadlocks on a saturated queue — pause admission, wait only for running executions


alekspetrov:
## Context

2026-08-03 v2.252.0 rollout: the release train published at 14:27Z and self-upgrade started at 14:28Z ("Waiting for 4 task(s) to complete..."). The drain waits for **all active executions (running + queued)** while the dispatcher **keeps admitting new work mid-drain** — #4677 was queued at 14:42Z, #4656 re-queued at 15:07Z, #4669 re-queued at 15:37Z, all after upgrade start. Two 30-minute drain windows timed out against a queue that never dropped below ~5; `self-upgrade failed` ERROR at 15:28Z. Operator had to manually install the binary and restart the daemon.

The failure compounded itself: the old binary's heartbeat bug (GH-4668, fixed by the very release being installed) kept churning tasks back into the queue with `infra` outcomes, guaranteeing the queue never emptied. A busy box structurally cannot self-upgrade.

## Acceptance

- While a self-upgrade is pending, the dispatcher stops starting new executions: queued rows stay queued, no new picks, pollers may continue enqueuing.
- The drain waits only for currently **running** executions (bounded by the existing timeout window); queued work survives the restart and resumes on the new binary.
- With 1 running + 5 queued tasks, the upgrade completes shortly after the running task finishes instead of timing out.
- The drain-timeout alert distinguishes "waiting on running task(s)" from the now-impossible "blocked by queue depth".

## Refs

- Incident: v2.252.0 rollout 2026-08-03, `daemon.log` 14:28–15:28Z on the founder box
- Compounding factor: GH-4668 heartbeat churn (its fix was aboard the blocked release)

alekspetrov:
PR #4688 was closed without merging: CI checks failed (Knowledge Graph Drift Gate); fix issue #4689 created to continue this work

This issue will not be retried automatically under its own number — see the reason above for what happens next.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4693 |
| Branch | `pilot/GH-4683` |
| Time to merge | 9m3s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4693 |
| Branch | `pilot/GH-4683` |
| Time to merge | 9m56s |

alekspetrov:
held for scope release schedule

Issue #4610, opened 2026-07-29
feat(autopilot): re-adopt held needs-manual-rebase PRs when their branch is updated — 5x recurrence in one wave


alekspetrov:
## Context

`escalateAndHold` (needs-manual-rebase) sets stage=failed, which is terminal in-memory: after an operator (or anyone) force-pushes a resolved rebase to the PR branch, autopilot never re-examines the PR. The only recovery is a fully manual `gh pr merge` (the external-merge scan then self-cleans tracking — that leg works).

Predicted in the 2026-07-28 close-out ("file a re-adopt-on-branch-update defect if this recurs often"). It recurred ×5 within one wave on 2026-07-29: pilot-console PRs #67, #68, #70, #74, #75 all parked needs-manual-rebase after sibling merges (parallel sub-issue PRs sharing main.go/config.go/go.mod), and every one needed operator rebase + manual merge (#67/#74 closed superseded; #68/#70/#75 rebased, CI green, merged manually).

## Fix

On push to a tracked-but-held PR branch (stage=failed with reason needs-manual-rebase, branch intact):
- re-enter the PR at `waiting_ci` (fresh CI on the new head), preserving attempt counters and the escalation audit trail;
- cap re-adoptions (e.g. 2) to avoid ping-ponging with a repeatedly-conflicting branch;
- the existing external-merge scan behavior stays as the fallback.

Detection can ride the existing PR poll (compare stored head SHA vs current) — no webhook needed.

## Why it matters

Multi-PR waves in one repo are now the normal S3+ dispatch shape; without re-adoption every sibling-merge cascade converts autopilot holds into serial operator merge work.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4612 |
| Branch | `pilot/GH-4610` |
| Time to merge | 8m38s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4612 |
| Branch | `pilot/GH-4610` |
| Time to merge | 9m32s |

alekspetrov:
held for scope release schedule

alekspetrov:
[...]

Issue #4798, opened 2026-08-07
fix(autopilot): platform-breaker metrics never reach AggregateMetrics.Snapshot — /metrics serves 0 (PR#4797 review)


alekspetrov:
## Problem

PR#4797 (GH-4791, platform-outage breaker part 1) added two fields to `internal/autopilot/metrics.go` — `Metrics.PlatformBreakerTrips` (counter) and `Metrics.PlatformBreakerOpen` (gauge) — and exports them as `pilot_platform_breaker_trips_total` / `pilot_platform_breaker_open` in `internal/gateway/prometheus.go`. The per-controller `Metrics.Snapshot()` copies both.

But `AggregateMetrics.Snapshot()` (`internal/autopilot/metrics_aggregate.go`) copies fields **explicitly, field-by-field**, and was not updated: it sums `CircuitBreakerTrips` (line ~119) and never touches the two new fields. Every real deployment reads `/metrics` through the aggregate (`SetMetricsSource(autopilotMetricsAggregate)`, `cmd/pilot/main.go`), so both new series serve 0 in production regardless of breaker state.

This is a recurrence of GH-4738 (fixed by PR#4739 in the same file on 2026-08-06): new snapshot fields silently dropped by the aggregate, invisible to tests that read a controller's own `Metrics`.

## Acceptance

1. In `AggregateMetrics.Snapshot()`: `agg.PlatformBreakerTrips += s.PlatformBreakerTrips` (each controller counts its own suppressed actions — sum is correct), and `agg.PlatformBreakerOpen = agg.PlatformBreakerOpen || s.PlatformBreakerOpen` (each controller mirrors the shared process-wide breaker state as of its own most recent `Observe`; OR is correct — any controller having last seen it open means open).
2. Composed regression test in the PR#4739 style: build ≥2 `*Metrics`, record a trip + set open on ONE of them, wrap in `NewAggregateMetrics`, assert the **aggregate** snapshot carries `PlatformBreakerTrips == 1` and `PlatformBreakerOpen == true`. This is the read path production actually uses.
3. `make build` / `make test` / `make lint` green.

## Scope fence

No changes to `PlatformBreaker` itself, `controller.go`, or the Prometheus exporter — the defect is confined to `metrics_aggregate.go` (+ its test file). The lazy-close wart noted on PR#4797 belongs to #4792, not here.

**This task must NOT be decomposed — implement as a single PR.** 

## Refs

[...]

PR #4806, opened 2026-08-07
feat(autopilot): platform-outage breaker part 2 — status probe, admission pause, held-PR re-drive

alekspetrov:
## Summary

Part 2 of 2 (TASK-458), building on part 1's cross-PR correlation and destructive-action suppression (#4791). Stops the daemon from spending during a platform outage and makes recovery automatic instead of operator-driven:

- **External status probe** (advisory, never gates open/close): `ProbeGitHubStatus` GETs GitHub's official component-scoped status API (`components.json` for the Actions component, `incidents/unresolved.json`) once on the breaker's open transition. 5s timeout, injectable getter for tests. Failure/parse-failure/staleness degrades to `unknown`. Deliberately does **not** use the third-party "OutageDeck" aggregator mentioned in an untrusted issue-thread comment.
- **Admission pause while open**: `Dispatcher.PauseAdmissionFor`/`ResumeAdmissionFor(owner string)` replace the old zero-arg pause with an owner-reference-counted set, so the platform breaker (`platform-breaker` owner) and the GH-4683 self-upgrade drain (`self-upgrade` owner) can share the same chokepoint without fighting over one bool. `Controller.SetAdmissionPauser` takes a narrow `AdmissionPauser` interface to avoid an `autopilot`→`executor` import cycle. Config-gated via `pause_admission` (default on).
- **Merge suppression while open**: `handleMerging` holds an already-green PR at `StageMerging` without merging — CI signal is untrustworthy platform-wide during an incident. Doesn't count against `MergeAttempts`.
- **Held-PR re-drive on close**: `handleCIFailed`'s suppression now parks PRs at `StageFailed` with a new `BreakerHoldActive` flag (mirrors the GH-4610 rebase-hold shape) instead of leaving them at `StageCIFailed` for indefinite reprocessing. A new periodic monitor (`startPlatformBreakerMonitor`) independently evaluates the breaker's time-based close condition (`EvaluateClose`) and re-drives every held PR back into `StageWaitingCI`, capped by `BreakerReadoptCount`.

Disabled by default; a nil `*PlatformBreaker` remains a byte-identical no-op.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/autopilot/... ./internal/executor/... ./cmd/pilot/... -race` all pass
- [x] Probe failure doesn't block breaker opening
- [x] Green probe doesn't veto correlation
- [x] Admission pause/resume correctly interleaves with the self-upgrade drain (owner-reference-counted)
- [x] Held PRs re-driven exactly once on close, capped re-adoption count respected
- [x] Merge suppressed while breaker open, resumes normally when nil/closed
- [x] Disabled-by-config paths verified as no-ops (nil breaker, `pause_admission: false`)


alekspetrov:
## Pre-merge review (Navigator, 2026-08-08) — APPROVE with follow-up filed

Strong implementation of the part-2 spec, and both review notes on #4792 were consumed correctly:

- **Periodic close evaluator**: `EvaluateClose()` extracted from `Observe`'s side effect (`closeIfQuietLocked`), driven by `startPlatformBreakerMonitor` on `probe_interval` — a held episode closes even with zero CI activity. Exactly the lazy-close fix requested.
- **Probe**: official component-scoped API only (`components.json` Actions component + `incidents/unresolved.json`), 5s timeouts, everything degrades to `unknown`, structurally advisory (runs strictly after `Observe` decided). Matches the endpoint guidance verbatim.
- **Owner-keyed admission pause** (`PauseAdmissionFor`/`ResumeAdmissionFor`) correctly prevents the breaker and the GH-4683 self-upgrade drain from fighting over one bool, with the drain's call sites migrated.
- Parking held PRs at `StageFailed`+`BreakerHoldActive` instead of `StageCIFailed` is the right call — per-tick reprocessing would refresh the quiet-period clock and the breaker could never time-close.

**Two confirmed defects, both in the recovery path — follow-up filed, not merge-blocking since the feature ships `enabled: false`:**

1. **Daemon restart mid-outage permanently strands held PRs.** `RestoreState` (`controller.go:1684`, untouched by this PR) skips every `StageFailed` row: `if pr.Stage == StageFailed { continue }`. A breaker-held PR survives restart in SQLite (`breaker_hold_active=1`) but never re-enters `activePRs` — and `ReDriveBreakerHeldPRs` iterates only `activePRs`. The new persistence columns cannot do their stated job ("hold survives a daemon restart"). A restart during a platform outage is not hypothetical — it's what happened on 2026-08-06.
2. **Observe-path close never re-drives.** `ReDriveBreakerHeldPRs` has exactly one caller: the monitor, gated on `if !breaker.IsOpen() { continue }` before its own `EvaluateClose`. When the close transition happens inside `Observe` (any CI failure landing after the quiet deadline), admission resumes and the close alert fires via `alertPlatformBreakerTransition` — but the monitor then sees the breaker already closed and skips, so held PRs stay parked until the re-adopt cap or a later episode that happens to close via the monitor.

Operator note: keep `platform_breaker.enabled: false` on the box until #4798 (metrics never reach the aggregate) and the recovery-path fix both merge.

Issue #4817, opened 2026-08-08
feat(executor): TASK-459 Phase 3 — recorded status is authoritative; no failure inferred from missing artifacts, no labels on closed issues


alekspetrov:
# feat(executor): TASK-459 Phase 3 — recorded status is authoritative; no failure inferred from missing artifacts, no labels on closed issues

## Context

Phase 1 (#4796 → PR#4802) landed the irreversible-action inventory and the `Verdict` contract. Phase 2 (#4811 → PR#4812) gated the CI-failure path. Phase 3 targets root pattern 2 from the task brief: **failure inferred from side-effects rather than read from recorded state** — the GH-4794 shape ("no PR appeared" read as failure while the ledger plainly said `superseded`) generalized to every remaining site — plus its sibling defect: label writes that ignore issue open-state (retry/progress labels stranded on closed issues).

The prior fix (`IsTerminalByDesignStatus`, `internal/executor/dispatcher.go:79`, GH-4794/#4800) covered one consumer family (`classifyWaitedExecution` → `EffectiveSuccess()`). This leg fixes the remaining instances in `cmd/pilot/` + `internal/executor/`, plus exactly one guard in `controller.go` whose own comment names the hazard.

Line refs verified on main @ `e3763efd` (2026-08-08). If drifted, re-trace from function names, not line numbers.

### 1. `recoverStaleRunningTasks` writes `stalled`, not `failed` — `internal/executor/dispatcher.go:528-644`

The write at :639 records `failed` after inferring purely from artifact absence: stale timestamp + no completed row (:561) + no live worker (:576) + no merged PR (:595) + no open PR (:621). That evidence is **liveness-loss, not failure** — and the boot-time sibling `reconcileOrphanedExecutions` already writes `ExecStatusStalled` for the same shape (:388, "orphaned by daemon restart"). Align: write `ExecStatusStalled` (via the same CAS-guarded call), keep every existing guard, the merged-PR heal (:595-613), and the GH-4423 open-PR liveness check unchanged. Downstream is already status-driven: `priorClaimWasStalled` (:1297) → stall carve-out (:1673) → `stallTaskAfterStallCap` at cap 8, instead of the failure ladder.

### 2. Stale-queued recovery status — decision point — `internal/executor/dispatcher.go:747-819`

[...]

Issue #4656, opened 2026-08-03
fix(executor): revalidate issue state at pickup and before PR creation — closed/superseded issues must not execute or open PRs


alekspetrov:
## Context

Blocked by: #4655

Same 2026-07-31 incident (record: `.agent/tasks/TASK-437-duplicate-execution-race-prevention.md`): child GH-4649's retry was claimed at 18:57 while legitimate, sat queued behind the serial ProjectWorker, and physically ran ~19:33→19:58 — by which time its scope had merged via the parent's PR#4652 (19:41) and its issue had been closed as superseded (19:43). The run completed anyway and opened PR#4653 against the closed issue at 19:58, born merge-conflicting.

Verified mechanism (refs at `d959cfd6`):

- Pickup-time re-checks (`internal/executor/dispatcher.go:2465-2528`) are strictly task_id-local: `hasTerminalSuccessLedger` + `decomposedChildrenAllComplete` read this task's own ledger rows. `buildTaskFromExecution` (`dispatcher.go:2551`) reconstructs the Task from the claim-time snapshot. **No code between claim and Execute re-fetches GitHub issue state.**
- `mergedPRPreflightCheck` (`dispatcher.go:662-675`) is keyed to `task.Branch` — a sibling/parent delivering the same scope on a different branch (`pilot/GH-4648`) is invisible to it by construction.
- Nothing gates PR creation on issue state: a finished run opens its PR regardless. `reconcileOrphanPRs` (`internal/autopilot/controller.go:5319-5373`) would even ADOPT such a PR into tracking — it has no issue-state awareness.
- `LabelSuperseded = "pilot-superseded"` (`internal/adapters/github/types.go:104`) has **zero functional callers** — the incident's issue-close was an executor LLM session improvising with `gh`, not a code path. This issue gives the system a real finalization path for that state; it does NOT add issue-closing logic.

## Acceptance

1. **Pickup-time revalidation** (`dispatcher.go` ~2465, alongside the existing re-checks): before `Execute()`, fetch live issue state for `exec.TaskID`. If the issue is CLOSED: do not execute — finalize the execution through `ExecutionLifecycle` with a typed terminal status (superseded/skipped vocabulary; no retry spawn, no new generation), journal an event naming the closed state and any `pilot-superseded` label found.
[...]

Issue #4101, opened 2026-07-08
fix(executor): stale-recovery status transitions must write execution_events


alekspetrov:
## Problem

`recoverStaleRunningTasks` / `recoverStaleQueuedTasks` (`internal/executor/dispatcher.go:319-328, 385`) mutate `executions.status` directly via `UpdateExecutionStatus` without writing an `execution_events` row. Any restart/orphan-driven terminal transition is therefore invisible in the event audit trail — a row killed by a restart reads identically to one still legitimately mid-flight.

Root-causing the 2026-07-08 GH-4050 duplicate-execution incident required cross-referencing raw `daemon.log` with store SQL and the exit-code classifier because `execution_events` for execution `5ce9bc2c` simply stops — no terminal event exists.

## Fix

Every status transition made by the stale-recovery paths writes a corresponding `execution_events` row (stage naming consistent with the existing event vocabulary; include the recovery reason in the event detail, e.g. `stale_running recovered after restart`).

Also audit the infra-classification path in the dispatcher worker (`internal/executor/dispatcher.go:977-991`) — an `infra` terminal status currently produces no event either; it should.

## Scope fence

- Observability only: no changes to WHICH transitions happen or their timing — only that they are recorded.
- Dashboard stage-ladder rendering (`buildStageInfo`, GH-4023 reducer) must not need changes; new events must map to existing rungs or resolve via the walk-back rule.

## Acceptance

- [ ] Stale-running and stale-queued recovery each emit an event on every row they touch
- [ ] Infra-classified terminal results emit a terminal event
- [ ] A test reconstructs a restart-orphaned execution's fate from `execution_events` alone
- [ ] Existing stage-strip regression tests unchanged and green

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4113 |
| Branch | `pilot/GH-4101` |
| Time to merge | 6m21s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4113 |
| Branch | `pilot/GH-4101` |
| Time to merge | 7m15s |

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4113 |
| Branch | `pilot/GH-4101` |
| Time to merge | 12m15s |

PR #4821, opened 2026-08-09
feat(executor): TASK-459 Phase 3 — recorded status is authoritative (GH-4817)

alekspetrov:
## Summary

TASK-459 Phase 3: migrates the executor/dispatcher/poller destructive-action sites named in the Phase-1 irreversible-action inventory to consult recorded ledger status instead of inferring failure from an absent side effect, and gates additive GitHub label/comment writes on the target issue's current open/closed state.

- `recoverStaleRunningTasks`/`recoverStaleQueuedTasks` (dispatcher.go) now write typed `stalled`/`canceled` instead of collapsing every boot-time orphan to `failed`.
- The GitLab (`handleGitlabIssueWithResult`) and CLI (`newGitHubRunCmd`) GH-3053 "no commit, no PR" demotions now consult a new `executor.IsNoArtifactExplainedOutcome` helper (`no_op`/`superseded`/`canceled`) before inferring failure from a missing commit/PR alone.
- Five additive label/comment-write sites now gate on a fresh `fetchIssueState` re-read and skip the write on positive evidence the issue is already closed, failing open on a lookup error:
  - `surfaceStalledIssue` (dispatcher.go)
  - `notifyTaskStartedSDK` caller (handlers.go, SDK-dispatch path)
  - `postTitleRejectionEscalation` (title_rejection.go)
  - `handleSubIssueCoverageGap` (epic.go)
  - `notifyExternalClose` label-correction block (controller.go) — reuses the already-fetched issue state; the informational comment is still posted unconditionally since it's pure history, not retry/label state.
- `checkWorktreePruned` (finish_tripwires.go) excludes `superseded`/`canceled` rows from its "commits but no PR" tripwire violation, alongside the pre-existing `decomposed` carve-out.
- `.agent/system/irreversible-actions.md` updated in place to record the new `re-read` evidence gates.

## Test plan

- [x] `go test ./internal/executor/... ./cmd/pilot/... ./internal/autopilot/...` — new regression tests per site (unit tests via `stubFetchIssueState` + fake `gh` CLI where directly testable; source-level guard tests via the existing `githubFuncBody` pattern for functions too network/SDK-entangled to unit test directly)
- [x] `make build` — green
- [x] `make test` — full suite green

### DIFF
--- a/.agent/system/irreversible-actions.md
+++ b/.agent/system/irreversible-actions.md
@@ -1,6 +1,6 @@
 # Irreversible-action inventory (TASK-459 Phase 1)
 
-**Status**: Phase 2 of 4 landed (Phase 5's false-success class split to TASK-460, 2026-08-08). Phase 1 built the inventory + `Verdict`
+**Status**: Phase 3 (GH-4817) landed 2026-08-09. Phase 1 built the inventory + `Verdict`
 contract with no behaviour change. Phase 2 migrated the CI-failure path —
 `handleCIFailed`'s ladder (family 1's `MaxCIFixIterations` close, family 3's
 pre-merge `CreateFailureIssue`, family 8's zero-evidence `escalateAndHold`)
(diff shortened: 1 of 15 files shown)


alekspetrov:
**Task 2 status decision** (documented here since `gh pr edit` isn't available to this session):

Chose `ExecStatusCanceled` (the doc-suggested bias) for `recoverStaleQueuedTasks` over keeping `failed`, mirroring `reclaimSelfOwnedQueuedChild`'s use of a typed administrative marker for the same non-outcome-write shape (a queued task that never ran because its project was removed / the daemon restarted mid-adoption).

Verified no downstream consumer depends on the literal `"failed"` status string for this reap path: `runner.go`'s `reapErrorSignatures`/`IsInfraNoise` match on the unchanged *reason string* (`"queued task orphaned by restart"`), not the status column, so the switch to `canceled` is transparent to that consumer. All existing dispatcher tests (updated to expect `canceled`) pass.

alekspetrov:
## Post-merge review (2026-08-10) — no defects found

Verified against the #4817 spec, all 8 tasks:

- **Tasks 1–2 (recovery writes)**: `stalled`/`canceled` via the same CAS-guarded `UpdateExecutionStatusIfNotTerminal`; `StageStalled`/`StageCanceled` events keep the GH-4101 audit trail. Wiring confirmed: `priorClaimWasStalled` (dispatcher.go:1337) reads `ExecStatusStalled`, so recovered rows take the stall carve-out → cap-8 path, not the failure ladder.
- **Tasks 3–4 (no-artifact demotions)**: `IsNoArtifactExplainedOutcome` = `no_op` ∪ `IsTerminalByDesignStatus`. GitLab site is inside `hr.Result != nil && hr.Result.Success`, so the new guard's `hr.Result.Outcome` deref is nil-safe. CLI path fails safe when `Outcome` is empty (helper returns false → behavior unchanged).
- **Task 5 (label gates)**: all five sites gate on positive closed-evidence only. `fetchIssueState`'s contract holds the fail-open invariant (never returns `Closed=true` except on a clean live read — GH-4656 acceptance #4). The one shadowing hazard checks out: in `notifyExternalClose`, `issue, err` from the :7337 fetch survive un-shadowed to the new gate (intermediate blocks use `cerr`/`searchErr`), and the `pilot-done` branch returns early, so the gate exclusively covers closed-without-done. Informational comment still posts unconditionally per spec. epic.go's gate sits after the ledger event + GitHub-parent guard — non-GitHub parents and the audit trail unaffected.
- **Task 6 (tripwire)**: `checkWorktreePruned` exclusion via `IsTerminalByDesignStatus`, decomposed carve-out retained.

Two notes, neither defect-worthy:

1. The `notifyExternalClose` closed-issue skip also suppresses the **removal** writes (`pilot-in-progress`/`pilot-failed` cleanup), so a closed issue can keep a stale `pilot-in-progress` label. Cosmetic — the poller skips non-open issues, and the spec declared removal-only writes harmless in either direction.
2. `gh4817_no_artifact_gate_test.go` asserts on handlers.go source text — brittle by nature, but follows the repo's existing `githubFuncBody` pattern for network-entangled functions.

Issue #4823, opened 2026-08-10
feat(autopilot): TASK-459 Phase 4 — one vocabulary per fact, destructive calls gated; drift and bypass become CI failures


alekspetrov:
# feat(autopilot): TASK-459 Phase 4 — one vocabulary per fact, destructive calls gated; drift and bypass become CI failures

## Context

Phases 1–3 landed the inventory + `Verdict` contract (#4796 → PR#4802), gated the CI-failure ladder (#4811 → PR#4812), and made recorded status authoritative in the executor/dispatcher (#4817 → PR#4821). Phase 4 is the final leg: make **drift and bypass impossible rather than unlikely** — every external fact has exactly one implementation, routing decisions consume typed values instead of substring-matching human-formatted strings, and a grep gate stops new call sites from bypassing the `Verdict` contract.

The first planned parity target is **already shipped**: GH-4779 landed the shared `ciFailureConclusions` table (`internal/autopilot/ci_monitor.go:667`) consumed by both `mapCheckStatus` and the evidence-gathering functions via `isCIFailureConclusion`, with anti-drift test `TestCIMonitor_MapCheckStatus_FailureConclusionParity` (`ci_monitor_test.go:589`). Do not redo it; the remaining targets are below.

Line refs verified on main @ `39881bba` (2026-08-10). If drifted, re-trace from function names, not line numbers. Full traces for tasks 3–4 are in `.agent/system/irreversible-actions.md` §9 — read that section before starting; do not re-derive.

Tasks are independent of each other — commit incrementally, do not entangle them.

### 1. Evidence-derivation parity test — `internal/autopilot/failure_class.go:280,314`

`ciFailureVerdictEvidence` (:314) re-derives each check's classification via `classifyCheckFailureFull` independently of `classifyPRFailure`'s aggregation (:280) — pattern-3 duplication. Its drift direction is conservative (mismatch → empty evidence → the Phase-2 gate reads it as evidence-free → hold), so this is a **test-only** task: add a fixture-corpus parity test asserting that for every corpus entry where `classifyPRFailure` returns non-`Unknown`, `ciFailureVerdictEvidence` returns non-empty. Reuse the existing #4787/TASK-418 fixture corpus in `failure_class_test.go` — do not invent a parallel corpus. No production code change.

### 2. Approval-source vocabulary: three implementations → one exported table

The channel/source vocabulary exists three times:

| Site | Form |
|---|---|
[...]

Issue #4831, opened 2026-08-11
fix(scripts): destructive-call gate Check 2 passes vacuously on macOS (BSD grep lacks -P); wire --self-test into pre-push


alekspetrov:
## Problem

Found in post-merge review of PR#4828 (TASK-459 Phase 4, issue #4823).

`scripts/check-destructive-calls.sh` Check 2 (bare `Verdict{}` composite literals) **passes vacuously on macOS**. Line 140:

```bash
hit=$(grep -PHn "$VERDICT_LITERAL_PATTERN" "$f" 2>/dev/null || true)
```

BSD grep (`/usr/bin/grep` on macOS) has no `-P`: it exits 2, stderr is discarded, and `|| true` converts the error into "no matches". Check 2 can never fire on macOS — the pre-push gate step [6/8] reports green regardless of violations. This is the exact "grep gate that can pass vacuously" class TASK-459 Phase 4 exists to eliminate, on the machine where pre-push runs.

The `--self-test` mode already detects this — but it is wired into nothing (not CI, not Makefile, not pre-push-gate.sh).

## Evidence (dev laptop, macOS)

```
$ bash scripts/check-destructive-calls.sh
✓ No destructive-call-gate bypasses found        # exit 0

$ bash scripts/check-destructive-calls.sh --self-test
  OK: seeded destructive-call violation detected
  FAIL: seeded Verdict{} violation was not detected
Self-test FAILED.                                 # exit 1
```

Failure scenario: a developer on macOS adds a bare `Verdict{}` in package `autopilot`; pre-push gate passes locally; the bypass is only caught at PR time by the Ubuntu CI leg.

## Tasks

1. **Make Check 2 portable.** Rewrite `scan_verdict_literals` / `VERDICT_LITERAL_PATTERN` (scripts/check-destructive-calls.sh:135-144) so it works under both BSD and GNU grep. Preferred: POSIX ERE `\bVerdict\{` match plus a post-filter that drops qualified matches (`.Verdict{`, `PreFlightVerdict{`, etc. — replacing the PCRE negative lookbehind `(?<!\.)`). Alternative floor: detect `grep -P` support once at startup and hard-fail with a clear message when unsupported — never swallow exit 2 into "clean".
2. **Wire the self-test.** Run `--self-test` at the start of step [6/8] in `scripts/pre-push-gate.sh` (currently lines ~191-199) and fail the step if it fails. This would have surfaced the bug on first local run.
[...]

PR #4834, opened 2026-08-11
fix(scripts): destructive-call gate Check 2 portable on BSD grep, wire --self-test into pre-push

alekspetrov:
## Summary

- `scan_verdict_literals` (Check 2, bare `Verdict{}` literal scan) used `grep -P` (PCRE), which BSD grep on macOS doesn't support — it exits 2, and the `|| true` on the caller silently turned that error into "no matches", so Check 2 passed vacuously on macOS while Linux CI correctly caught violations.
- Replaced the PCRE negative-lookbehind pattern (`(?<!\.)\bVerdict\{`) with a portable POSIX ERE (`(^|[^A-Za-z0-9_.])Verdict\{`, `grep -E`) that gets the same two exclusions (qualified idents like `sdkcore.Verdict{}`, embedded-suffix names like `PreFlightVerdict{}`) via a preceding-character class instead of lookbehind/`\b`.
- Wired `--self-test` into `pre-push-gate.sh` step [6/8] so a broken detector fails the gate instead of reporting false-green — this is exactly what would have caught the bug on the dev laptop that found it.
- Allowlist semantics unchanged (`VERDICT_LITERAL_ALLOWLIST`, Check 1's commented allowlist).

Found in post-merge review of PR#4828 (TASK-459 Phase 4, #4823).

## Test plan

- [x] `bash scripts/check-destructive-calls.sh --self-test` passes (Linux/GNU grep in this sandbox)
- [x] `bash scripts/check-destructive-calls.sh` plain run passes clean on current `main`
- [x] `scripts/pre-push-gate.sh` step [6/8] now runs `--self-test` before the plain scan and fails the step on self-test failure
- [x] CI job `destructive-call-gate` (`.github/workflows/ci.yml`) unchanged — still runs the plain script on Ubuntu, stays green

### DIFF
--- a/scripts/check-destructive-calls.sh
+++ b/scripts/check-destructive-calls.sh
@@ -77,12 +77,25 @@ DESTRUCTIVE_CALL_ALLOWLIST=(
 # owning file.
 # ---------------------------------------------------------------------------
 #
-# Uses a negative lookbehind so a *qualified* identifier of the same name in
-# an unrelated package (e.g. cmd/pilot/poller_github.go's sdkcore.Verdict)
-# doesn't false-positive, and a leading \b so embedded-suffix type names
-# (PreFlightVerdict{}, JudgeVerdict{} in internal/executor/intent_judge.go)
-# don't either.
-VERDICT_LITERAL_PATTERN='(?<!\.)\bVerdict\{'
+# POSIX ERE (grep -E), portable across BSD grep (macOS) and GNU grep
+# (Linux CI) — deliberately NOT `grep -P`. BSD grep has no -P: it exits 2
+# on an unsupported pattern, and the caller's `|| true` (needed to keep
+# "no matches" from tripping `set -e`) silently turns that error into an
+# empty, "clean" result. That let this exact check pass vacuously on macOS
+# pre-push while flagging real violations in Linux CI (found in post-merge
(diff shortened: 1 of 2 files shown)


alekspetrov:
## Post-merge review (2026-08-11)

**Verdict: clean — verified empirically on the machine that had the bug.**

- Live run on the macOS/BSD-grep laptop that surfaced GH-4831: `--self-test` now passes all four checks (seeded destructive call caught · seeded `Verdict{}` caught with qualified/suffix look-alikes ignored · both allowlisted files unflagged); plain run clean on main.
- Pattern equivalence verified: `(^|[^A-Za-z0-9_.])Verdict\{` reproduces both PCRE exclusions — `.` in the class rejects `sdkcore.Verdict{`, alnum/underscore rejects `PreFlightVerdict{`/`JudgeVerdict{` — with underscore parity matching the original `\b` semantics.
- Self-test wired ahead of the plain scan in pre-push step [6/8], failure increments the gate's FAILURES — exactly the "would have caught it on first local run" placement.
- Allowlists untouched; CI job untouched (was already sound on GNU grep).

Note (not defect-worthy): `2>/dev/null || true` still swallows any *other* grep error class (e.g. unreadable file) — the per-machine self-test is the designed guard for that, which is acceptable.

Issue #4748, opened 2026-08-06
feat(gateway): C14 pilot leg — approval read surface + POST decision endpoint on the DecisionRecorder seam


alekspetrov:
# feat(gateway): C14 pilot leg — approval read surface + POST decision endpoint on the DecisionRecorder seam

## Problem

S4 C14 ("Needs You") requires the console to see pending approvals and record decisions. Today neither is possible: the gateway's `/api/v1` route table (`internal/gateway/server.go:231-238`) is GET-only with no approval data in any response, and `approval_request_id` — the key `DecisionRecorder` decisions are recorded against — is exposed on no HTTP surface. This is the "pilot (1 endpoint)" leg the roadmap §S4 row names (it is actually one read route + one write route).

All facts verified at origin/main 2026-08-06 (research pass).

## Context (verified)

- **Seam**: `DecisionRecorder` (`internal/approval/types.go:80-82`) — `RecordDecision(ctx, requestID string, decision Decision, by string) error`. Implemented by `*approval.Manager.RecordDecision` (`internal/approval/manager.go:412-436`) → `memory.Store.SetApprovalDecision` (`internal/memory/store.go:1472-1496`), keyed on `executions.approval_request_id`. Wired today only to Slack/Telegram handlers (`WithDecisionRecorder`, call sites `cmd/pilot/main.go:609,636,1697,1718`, `internal/pilot/pilot.go:236`).
- Restart-survival of pending approvals is real and integration-tested (`internal/autopilot/restart_approval_*_integration_test.go`) — the rehydration path has a pending-approvals predicate (grep log msg `"rehydrated pending approvals"`); reuse that predicate, do not invent a new one.
- `handleAutopilot` (`server.go:555`) already serves PR `stage` incl. `awaiting_approval`, but not `approval_request_id`, and the console proxy can't reach it (console-side allowlist — NOT this task's problem).
- The gateway has an `Authenticator` — **verify how sibling `/api/v1` routes are wrapped before wiring; the new routes must be authed identically** (the decision route is a state-changing action; it must never be weaker-authed than the rest of the surface).

## Acceptance

1. **Read**: `GET /api/v1/approvals` → pending approval requests, JSON array of `{requestId, executionId, taskId, projectPath, prNumber, prUrl, requestedAt}` (fields nullable where the ledger lacks them). Pending = the same predicate the approval rehydration path uses. Project-scoped by the same mechanism as sibling dashboard routes (`dashboardProjectPath` filter where applicable).
[...]

Issue #4431, opened 2026-07-17
fix(slack): approval buttons unroutable in Socket Mode — HandleInteraction only wired to HTTP webhook; clicks fall through to comms 'No pending task to confirm' (#4411 root cause, half 2)


alekspetrov:
## Symptom

Every Slack approval click today returned "No pending task to confirm" — including clicks on messages posted AFTER the #4426 persistence fix was deployed (evidence: approval for PR #4429 posted 21:38Z on v2.241.2-9, clicked minutes later, same error; no restart in between).

## Root cause (confirmed in code)

`approval.SlackHandler.HandleInteraction` is only invoked from the HTTP interactivity webhook:

- `internal/pilot/pilot.go` (~line 266): `p.slackInteractionWH = slack.NewInteractionHandler(signingSecret)` + `OnAction(... HandleInteraction ...)` — an HTTP endpoint requiring a public Interactivity Request URL.
- The founder deployment runs Slack in **Socket Mode** with no public HTTP endpoint, so that webhook never receives anything.
- Socket-mode button clicks arrive as `core.MessageEvent{Action:"callback"}` → `internal/adapters/slack/handler.go HandleMessage/HandleCallback` → `comms.Handler` → `handleConfirmation` → `pendingTasks[contextID]` miss → "No pending task to confirm." (comms/handler.go:705).

So approval buttons have never been routable on socket-mode deployments. #4426 (rehydrate/persistence) is necessary but insufficient — the interaction never reaches the approval handler.

## Fix

In the socket-mode callback path, route approval interactions to `slackApprovalHdlr.HandleInteraction` BEFORE the comms fallthrough: match the approval buttons' action IDs (or `value` prefix `approve:`/`reject:`) in `HandleMessage`'s `"callback"` case / `HandleCallback`, and only fall through to comms for execute/cancel task actions. Mirror how Telegram routes its approval callbacks (GH-3825 precedent). Also: the final fallthrough reply should distinguish "unknown action" from "no pending task".

## Acceptance

- Socket-mode deployment: click Merge on a fresh Pre-Merge Approval message → PR merges (or decision recorded) — no HTTP webhook configured.
- Click after a daemon restart on a pre-restart message → works (#4426's rehydrate covers this half).
- execute/cancel task buttons still reach comms.

## Refs

#4411 (symptom issue), PR #4426 (persistence half), GH-3825 (Telegram routing precedent). Founder-priority: pilot reliability tier.

[...]

Issue #4411, opened 2026-07-17
fix(approval): pending approval requests are in-memory only — daemon restart kills every posted Merge button ('No pending task to confirm'), PRs strand at awaiting_approval


alekspetrov:
## Symptom

Founder clicked **Merge** on a Slack approval message for an `awaiting_approval` PR (#4408) and got:

> No pending task to confirm.

## Root cause

Every pending-approval structure is an in-memory map, none persisted:

- `internal/approval/slack.go` — `SlackHandler.pending map[string]*slackPending` (button value carries `approve:<requestID>`; requestID lookup fails after restart)
- `internal/approval/manager.go` — `Manager.pending map[string]*pendingRequest` (only *decisions* persist via `PRStateWriter`; requests don't)
- `internal/approval/github.go` — `GitHubHandler.pending` (same; also means a `gh pr review --approve` after restart may have no poller waiting on it)
- `internal/comms/handler.go:697` — `pendingTasks` map; the unmatched click falls through here, producing the misleading "No pending task to confirm." reply

Daemon restart (12:08Z today, #4403 deploy) wiped all of them. Every approval button posted before a restart is permanently dead, while the PR sits `awaiting_approval` in `autopilot_pr_state` — the ledger says waiting, but nothing is listening.

## Expected

Approval requests must survive restart. Either:

1. **Rehydrate**: persist pending requests (they're nearly there — `autopilot_pr_state.stage='awaiting_approval'` already encodes the need) and on boot re-arm handlers for every awaiting_approval PR, honoring old buttons' requestIDs; or
2. **Re-post**: on boot, for each `awaiting_approval` PR, invalidate the stale Slack message (edit it to "expired — see new message") and post a fresh approval request.

Option 2 is simpler and self-healing. Also: route an unmatched `approve:<id>` action to a clear "this approval expired (daemon restarted); a new request will be posted" reply instead of the comms-handler fallthrough.

## Refs

- Incident: #4408/#4410 unapprovable from Slack post-restart, 2026-07-17 ~12:45Z, box v2.241.1-7-g868e8fd1
- Related: TASK-404 (lifecycle chokepoint — same "in-memory state vs restart" class)

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4426 |
| Branch | `pilot/GH-4411` |
| Time to merge | 54m22s |

alekspetrov:
held for scope release schedule

Issue #4835, opened 2026-08-11
feat(gateway): C17 pilot leg — operator chat API, web transport for the comms adapter brain


alekspetrov:
## Problem

The approved console v4 dashboard includes an **Operator chat panel** — conversational control of the tenant instance from the web console. The brain already exists and is shared: `internal/comms` (GH-2143) — Telegram, Slack, and Discord all feed `comms.Handler.HandleMessage` and reply through the `comms.Messenger` interface. What's missing is a web transport: an HTTP API on the gateway that the console control plane can proxy to. This issue adds that transport as a fourth `BuildHandler` assembly. Console-side proxy is a separate issue in qf-studio/pilot-console.

## Context (verified @ main, 2026-08-11)

- Inbound seam: `comms.Handler.HandleMessage(ctx, *comms.IncomingMessage)` — `internal/comms/handler.go:162`. `IncomingMessage` (`internal/comms/types.go:10-25`) carries `ContextID`, `SenderID`, `SenderName`, `Text`, `Platform`, `IsCallback`/`CallbackID`/`ActionID` (button presses), `Timestamp`.
- Outbound seam: `comms.Messenger` — `types.go:28-51`: `SendText`, `SendConfirmation` (returns messageRef), `SendProgress(contextID, messageRef, taskID, phase, progress, detail)` (edit-in-place by ref), `SendResult(taskID, success, output, prURL)`, `SendChunked`, `AcknowledgeCallback`, `MaxMessageLength`. Existing impls: `internal/adapters/telegram/messenger.go:12`, `internal/adapters/slack/messenger.go:11`.
- Assembly point: `comms.BuildHandler(HandlerDeps)` — `internal/comms/factory.go:121`; existing call sites `cmd/pilot/main.go:2782` (Telegram, TaskIDPrefix "TG"), `main.go:3532` (Slack), `cmd/pilot/poller_discord.go:65`.
- Gateway: routes register inline in `Server.Start` (`internal/gateway/server.go:216-267`); protected routes go on `apiMux` behind bearer middleware (`server.go:247-251`; auth = GH-4784, `internal/gateway/auth.go:58-69`, enabled via config `auth.type: api-token`). Read-endpoint pattern to copy: #4749 `internal/gateway/events.go` + `events_test.go`. POST pattern: `handleApprovalDecision` `internal/gateway/approvals.go:170`.
[...]

PR #4838, opened 2026-08-11
feat(gateway): C17 pilot leg — operator chat API, web transport for the comms adapter brain

alekspetrov:
## Summary
- Adds `internal/adapters/web` — a fourth `comms.BuildHandler` transport backing a web HTTP chat surface for the console dashboard's Operator chat panel, reusing the same `comms.Handler.HandleMessage` brain Telegram/Slack/Discord already use.
- `WebMessenger` implements `comms.Messenger` via a per-conversation, monotonically seq-numbered in-memory event buffer (`text`/`confirmation`/`progress`/`result`), 500-event cap (drop-oldest) + 1h inactivity expiry, with `messageRef` sharing so progress events can be grouped/replaced in place.
- Gateway routes on `apiMux` (bearer auth inherited): `POST /api/v1/chat/messages` (dispatches on the daemon context, not the request context, so a task-shaped message can run for minutes without blocking the handler) and `GET /api/v1/chat/conversations/{id}/events?after=<seq>`. Registered only when `adapters.chat.enabled` is true.
- Approve/reject-shaped callbacks are rejected with 400 — approvals stay on `POST /api/v1/approvals/{requestId}/decision` (#4748), matching the GH-4411/GH-4431 scope fence; the chat brain has no approval semantics.
- Wired at both gateway construction sites (`cmd/pilot/main.go` polling mode, `internal/pilot/pilot.go` gateway mode) so the API exists in either daemon mode.
- New `adapters.chat` config block (disabled by default) documented in `configs/pilot.example.yaml`.

## Test plan
- [x] `internal/adapters/web`: unit tests for `WebMessenger` (seq ordering, buffer cap drop-oldest, 1h expiry, messageRef sharing/reuse, PRUrl/Success) and `API` (validation errors, approval-callback rejection, accept-time seq, daemon-ctx dispatch).
- [x] `internal/gateway/chat_test.go`: unit-level handler tests (405/400/503/202/200) plus two composed real-HTTP-listener tests — routes absent (404) when `adapters.chat` is disabled, and a full auth + operational-query + task confirmation → execute → progress/result flow driven by a fake `executor.Backend`.
- [x] `go build ./...`
- [x] `go test -race ./...` (all packages pass)
- [x] `make lint` (golangci-lint not installed in this sandbox; command exits 0)

### DIFF
--- a/.agent/system/FEATURE-MATRIX.md
+++ b/.agent/system/FEATURE-MATRIX.md
@@ -140,6 +140,7 @@
 | Discord adapter wiring | ✅ | main | `--discord` | `adapters.discord` | Wire Discord poller, config, and CLI flag in main.go (v2.30.0, PR #1882) |
 | Asana CompleteTask callback | ✅ | adapters/asana | - | - | Wire Asana CompleteTask on successful PR creation (v2.10.0, PR #1720) |
 | Telegram memory store | ✅ | adapters/telegram | - | - | Wire memory store to Telegram HandlerConfig (v2.25.0, PR #1754) |
+| Web chat API | ✅ | adapters/web, gateway | - | `adapters.chat` | Operator chat HTTP transport for console dashboard: `POST /api/v1/chat/messages` + `GET /api/v1/chat/conversations/{id}/events`, backed by comms.Handler via a per-conversation seq-numbered in-memory event buffer (500-event cap, 1h expiry); dispatches on the daemon context so tasks survive request cancellation; rejects approve/reject-shaped callbacks (approvals route through `/api/v1/approvals/{requestId}/decision` instead) (GH-4835) |
 
 ## Output/Notifications
 
--- a/.agent/tasks/gh-4835.md
+++ b/.agent/tasks/gh-4835.md
@@ -1,0 +1,87 @@
+# GH-4835
+
+**Created:** 2026-08-11
+
+## Problem
+
+GitHub Issue GH-4835: feat(gateway): C17 pilot leg — operator chat API, web transport for the comms adapter brain
+
+## Problem
+
+The approved console v4 dashboard includes an **Operator chat panel** — conversational control of the tenant instance from the web console. The brain already exists and is shared: `internal/comms` (GH-2143) — Telegram, Slack, and Discord all feed `comms.Handler.HandleMessage` and reply through the `comms.Messenger` interface. What's missing is a web transport: an HTTP API on the gateway that the console control plane can proxy to. This issue adds that transport as a fourth `BuildHandler` assembly. Console-side proxy is a separate issue in qf-studio/pilot-console.
+
+## Context (verified @ main, 2026-08-11)
+
+- Inbound seam: `comms.Handler.HandleMessage(ctx, *comms.IncomingMessage)` — `internal/comms/handler.go:162`. `IncomingMessage` (`internal/comms/types.go:10-25`) carries `ContextID`, `SenderID`, `SenderName`, `Text`, `Platform`, `IsCallback`/`CallbackID`/`ActionID` (button presses), `Timestamp`.
+- Outbound seam: `comms.Messenger` — `types.go:28-51`: `SendText`, `SendConfirmation` (returns messageRef), `SendProgress(contextID, messageRef, taskID, phase, progress, detail)` (edit-in-place by ref), `SendResult(taskID, success, output, prURL)`, `SendChunked`, `AcknowledgeCallback`, `MaxMessageLength`. Existing impls: `internal/adapters/telegram/messenger.go:12`, `internal/adapters/slack/messenger.go:11`.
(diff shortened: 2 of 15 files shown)


alekspetrov:
## Post-merge review — C17 pilot leg (TASK-463 / #4835)

**Verdict: contract substantively honored — approved as merged, with one wiring defect needing a follow-up fix and one contract gap console#115 must anchor around.**

### Contract checklist

- ✓ **Poll-drain, not SSE/streaming** — `POST /api/v1/chat/messages` returns 202 immediately; `Dispatch` spawns `HandleMessage` on the daemon context (`internal/gateway/server.go:219-223` stashes `Start`'s ctx; main.go:2642 / pilot.go:866 both pass daemon-lifetime contexts). `GET .../events` drains and returns; no handler holds the response open against the 15s WriteTimeout.
- ✓ **Shared brain reused** — `comms.BuildHandler` with a full 7-method `comms.Messenger` impl (`var _ comms.Messenger = (*WebMessenger)(nil)`), Platform `web`, ContextID `web:<convID>`. `web.API.Dispatch` only validates and forwards to `HandleMessage` — no brain fork. New assemblies at `cmd/pilot/main.go:2618` (polling) and `internal/pilot/pilot.go:822` (gateway), the 4th platform on the seam.
- ✓ **Approvals never through the brain** — `isApprovalCallback` (`internal/adapters/web/api.go:52`) 400s actionId `approve`/`reject` and callbackId `approve:`/`reject:` prefixes; comms itself rejects unknown actionIds (GH-4431 fix) as a second fence. No approval verbs on the surface, no side-channel. Tested at unit + composed level.
- ✗→✓ **Both construction sites wired — but gateway mode has a nil-runner hole** (defect 1 below). Polling mode is correct (`SetChatAPI` before `Start`, main.go:2611-2630).
- ✓ **Bearer auth** — both routes on `apiMux` behind `s.auth.Middleware` (`server.go:255-271`); WS bypass not extended; composed test asserts 401 on both routes.
- ✓ (mostly) **Tests** — seq ordering + after-cursor, 202 semantics, auth rejection, buffer cap, 1h expiry, disabled⇒404, approval fence all covered. Gap: defect 4.

### Defects

1. **[Medium-High] Chat API silently absent in gateway mode when no polling adapter is enabled.** `needsPollingInfra` (`cmd/pilot/main.go:663`) doesn't include `cfg.Adapters.Chat.Enabled`, so `gwRunner` stays nil; `WithChatHandler(gwRunner, ...)` at `cmd/pilot/main.go:1072` passes nil; the guard at `internal/pilot/pilot.go:815` (`p.chatRunner != nil`) then skips `SetChatAPI` entirely — routes 404 while the log at `main.go:1073` claims "Chat API enabled in gateway mode". A gateway-mode daemon serving only webhooks + the console chat panel hits this. Exactly the "silently doesn't exist in that mode" class the spec called out, just in config-space rather than code-space. Fix: add chat to `needsPollingInfra` (or construct a runner in the chat branch), and only log after wiring actually succeeds.
2. **[Medium] Buffer-reset detection is not implementable by a polling client.** The spec/docs (`internal/adapters/web/messenger.go:571-579` Event doc) require clients to treat `after > newest seq` as a reset — but `GET .../events` returns only `{conversationId, events}`, never the newest seq, and events with `seq <= after` are filtered out. After daemon restart or 1h expiry, a client with a stale cursor polls into silence, then silently loses events `1..cursor` once the new numbering passes the cursor. `internal/adapters/web/api.go:124` computes `latestSeq` and discards it (`events, _ :=`). Fix: add `latestSeq` to `chatEventsResponse` (`internal/gateway/chat.go:120-124`). Same class applies to drop-oldest gaps past the 500 cap.
3. **[Low] `events` serializes as `null`, not `[]`, for unknown/expired conversations.** `messenger.go:241` returns nil for a missing buffer; `handleChatEvents` writes it verbatim. A known-but-caught-up conversation returns `[]`. Spec says "empty array when nothing new"; JS clients doing `resp.events.length` break on the null case.
4. **[Low] The spec's request-context-cancellation test is not actually implemented.** `TestAPI_Dispatch_UsesGivenContextNotCancelledOne` (`internal/adapters/web/api_test.go:429`) never cancels anything — it passes `context.Background()` and its own comment concedes it only simulates a correctly-wired caller. The gateway test asserts ctx identity with `daemonCtx`, which verifies wiring but not the survival behavior.
5. **[Info] Per-conversation rate limit is rotatable** — a bearer holder can spread messages across fresh conversationIDs; conversation count between lazy prunes is unbounded (any messenger call sweeps all conversations, so bounded in practice). No `MaxBytesReader` on the POST body (pre-existing gateway convention). All bearer-gated, operator-facing — acceptable, noting for the record.
6. **[Info] conversationIDs containing `/` can be created via POST but never polled** (GET `{id}` matches a single path segment). No charset validation on conversationID.

Event-buffer concurrency is clean: single mutex on all paths, `Events` copies into a fresh slice, shared `Progress`/`Success` pointers are never mutated after append; cap reslice keeps backing-array growth bounded. `go test -race` green.

Issue #4841, opened 2026-08-11
fix(autopilot): recovery-owner designation must survive restart — persist it and extend the seam to the review path


alekspetrov:
# fix(autopilot): recovery-owner designation must survive restart — persist it and extend the seam to the review path

**Status**: 📋 Ready to dispatch
**Created**: 2026-08-11
**Assignee**: Pilot

## Context

PR#4840 (GH-4826) built the right seam: `spawnFailureIssue` (internal/autopilot/controller.go:2462) designates exactly one recovery owner per CI-failure close, and TerminalLabel is set iff `CreateFailureIssue` returns a live issue. Post-merge review found the designation is **in-memory only** — one daemon restart in the wrong window resurrects the exact double-arm incident the PR fixed. The daemon has a documented history of mid-flow restarts (GH-4807).

Crash window: `TerminalLabel` (internal/autopilot/types.go:1169) has **no column** in `SavePRState`/`GetPRState` (internal/autopilot/state_store.go:638-727). A crash between the PR close (controller.go:2720) and the end-of-ProcessPR persist (controller.go:2093) rehydrates the PR label-less at StageCIFailed; `checkExternalMergeOrClose` (processAllPRs, controller.go:6881) runs before any handler, has no StageCIFailed guard, and the self-close markers (controller.go:582-591) are also in-memory — so `notifyExternalClose` (controller.go:7455-7458) defaults to `pilot-retry-ready` while the spawned fix issue is live. Double-arm, restored by restart.

Same shape on the review path: `handleReviewRequested` still closes before designating — direct `CreateReviewIssue` (controller.go:3007), close (controller.go:3050), designation after (controller.go:3065), in-memory.

## Implementation

1. **Persist the ownership decision.** Either add a `terminal_label` column to the PR-state store (SavePRState/GetPRState/RestoreState roundtrip, migration included) **or** make `notifyExternalClose` consult the durable `autopilot_spawned_fixes` claim before defaulting to retry-ready. Prefer whichever is smaller and closes BOTH the pre-merge (controller.go:2672) and post-merge (controller.go:4283) rungs; state your choice in the PR description.
2. **Extend the designate-before-close seam to `handleReviewRequested`** — route it through `spawnFailureIssue`-style ordering (designate, then close), or persist its designation the same way as (1).
[...]

Issue #4826, opened 2026-08-10
fix(autopilot): CI-failure close arms both fix-issue and source-retry chains — no designated recovery owner


alekspetrov:
## Context

After the CI-failure close of PR#4818 (2026-08-09), BOTH recovery paths armed simultaneously:

- the fix-issue chain (#4820, spawned by `CreateFailureIssue`), and
- the source-issue retry chain (#4817 unmarked → retry generation → PR#4821).

GH-3806's `TerminalLabel` mechanism exists precisely to designate a follow-up issue as the retry owner (mark the source terminal so it isn't re-queued while a follow-up owns the work), but it didn't fire on this path — the close handler spawned the fix issue without marking the source, so two executors could have raced the same failure. Benign this time only because preflight `reject_vague`-ed #4820; a fix-issue vs source-retry pair has different task keys, so TASK-407's admission claim cannot catch it (that guard is same-task only).

Anchors (@ main 2026-08-11):

- `internal/autopilot/controller.go:2419` — `handleCIFailed` ("creates fix issue via feedback loop"); spawn at `:2637` → `feedbackLoop.CreateFailureIssue`.
- `prState.TerminalLabel` set sites: `controller.go:2581, 2707, 2964, 3030` (LabelFailed), `:5375` (LabelSuperseded); poller-side respect at `:7269-7288`.
- GH-3806: the designating mechanism this path failed to invoke.

## Implementation

1. Map the CI-failure close ladder in `handleCIFailed` + its callees: for each branch that reaches `CreateFailureIssue`, determine what currently marks (or fails to mark) the source issue/PR terminal, and which branches leave the source eligible for retry re-queue.
2. Enforce single ownership at the spawn point: when `CreateFailureIssue` succeeds, the source must be marked terminal (fix issue owns recovery); when no fix issue is spawned, the source retry chain owns it. The choice may stay class-dependent (infra vs code per the TASK-459 verdicts) — but **simultaneous arming must be structurally impossible at the call site**, not dependent on preflight rejecting the fix issue.
3. Wire the invariant where the spawn happens (one seam), not per-branch — a new branch added later must inherit exclusivity by construction.

## Acceptance

- [ ] Test reproducing the #4818 shape: CI-failure close spawns a fix issue → source issue is marked terminal and the retry path declines to re-queue it (assert through recorded state, not a mock echo).
- [ ] Test: failure class where no fix issue is spawned → source retry chain remains armed; no terminal mark.
[...]

Issue #4842, opened 2026-08-11
fix(autopilot): owner-death recovery — a designated fix issue that dies must re-arm the source or escalate


alekspetrov:
# fix(autopilot): owner-death recovery — a designated fix issue that dies must re-arm the source or escalate

**Status**: 📋 Ready to dispatch
**Created**: 2026-08-11
**Assignee**: Pilot

## Context

After PR#4840 (GH-4826), a CI-failure close designates exactly one recovery owner — usually the spawned fix issue. Post-merge review found there is **no handling for the owner dying**: the GH-4459 comment (internal/autopilot/controller.go:2673) claims the fix issue "cleared preflight admission," but `CreateFailureIssue` (internal/autopilot/feedback_loop.go:221-325) only creates the issue — the preflight judge (`reject_vague`, internal/executor/intent_judge.go) runs much later. A designated fix issue rejected at preflight leaves **zero owners**: the PR is closed, retry is suppressed by the designation, nothing ships. In the original GH-4820 incident it was the (then-buggy) second arm that actually delivered the fix — with exclusivity now enforced, that accidental safety net is gone.

Adjacent hole: the dedup path returns a previously recorded fix issue with **no open-state check** (feedback_loop.go:255-263; state_store.go:1191) — a closed/dead fix issue can be re-designated owner.

## Implementation

1. **Detect owner death.** A designated fix issue that (a) is declined by the preflight judge, or (b) is closed without a merged PR, is a dead owner. Detection may hook the decline path (the judge already records `declined-preflight` executions) and/or a periodic scan of designated-owner issues' open/merged state — pick the seam that needs no new poller if one exists.
2. **React**: on owner death, either re-arm the source (clear the designation so the normal retry ladder applies, respecting retry generations) or escalate `needs-human` when retries are exhausted — never silently strand. Log + alert through the existing alerts engine.
3. **Dedup open-state check**: before returning a deduped fix issue as owner (feedback_loop.go:255-263), verify it is still open; if closed-unmerged, treat as dead and fall through to creation or escalation.
4. **Tests**: table-driven owner-death matrix — preflight-declined / closed-unmerged / closed-merged (not death) / dedup-returns-closed — asserting exactly one live owner or an escalation in every row.

Out of scope: persistence of the designation itself (separate issue, dispatched alongside); changes to the preflight judge's decision logic.

## Acceptance

[...]

Issue #3903, opened 2026-07-06
fix(autopilot): scope PR state per repo — cross-repo PR-number collisions drive wrong-repo actions


alekspetrov:
## Problem

`autopilot_pr_state` and `autopilot_pr_failures` key rows on `pr_number` alone — no repo column. With per-project autopilot controllers (one per `projects:` repo + the default repo), every controller restores EVERY row at startup: `StateStore.LoadAllPRStates()` (`internal/autopilot/state_store.go:372`) has no repo filter, and `Controller.RestoreState()` (`internal/autopilot/controller.go:~507`) adopts all non-failed rows into its `activePRs`, then processes those PR numbers against its own `c.owner/c.repo`.

## Observed incident (2026-07-06, daemon v2.214.1)

Startup log — all 8 controllers restored the same 16 rows:

```
10:32:16 msg="Restored autopilot PR states from SQLite" repo=qf-studio/auth-service count=16
10:32:16 msg="Restored autopilot PR states from SQLite" repo=qf-studio/studio-sdk count=16
10:32:16 msg="Restored autopilot PR states from SQLite" repo=alekspetrov/ai-coding-summit count=16
10:32:16 msg="Restored autopilot PR states from SQLite" repo=qf-studio/pilot-canary-sandbox count=16
10:32:16 msg="Restored autopilot PR states from SQLite" repo=qf-studio/pilot count=16
```

Rows for PRs #74/#75 belonged to `qf-studio/studio-sdk` (merged 2026-07-05). After restart, each controller acted on those numbers against its own repo:

- studio-sdk controller: correctly resolved "PR merged externally" — fine.
- **qf-studio/pilot controller: fetched pilot's ancient PR #74 (closed unmerged in Feb) → "PR closed externally without merge" → applied `pilot-retry-ready` to pilot issues #72/#73** (closed since 2026-01-28; labels manually removed on 2026-07-06) **and deleted branches `pilot/GH-72`/`pilot/GH-73` in the wrong repo.**
- Controllers for repos where #74/#75 don't exist as PRs: `GetPullRequest` returns 404 forever; the 404 path in `processAllPRs` just `continue`s, so the rows never resolve → infinite retry loop (`failed to fetch PR` ≈2/s, sustained for hours, burning API quota).

Risk class: any cross-repo PR-number collision lets a controller mutate the wrong repo's PRs/issues/branches — labels, issue close, branch delete, and potentially merge processing when the colliding number is an open PR in the wrong repo.

[...]

PR #4846, opened 2026-08-11
fix(autopilot): recovery-owner designation survives restart, extended to review path (GH-4841)

alekspetrov:
## Summary

- Persists the recovery-owner designation (`prState.TerminalLabel`) across daemon restarts by making `notifyExternalClose` fall back to a durable `autopilot_spawned_fixes` claim (via new `HasSpawnedFixForPR`) when the in-memory label was lost to a crash between PR close and `persistPRState`. No schema migration needed — the claim row is written synchronously by `CreateFailureIssue`/`CreateReviewIssue` before either handler closes the PR, so it inherently survives the crash window that a `terminal_label` column on `autopilot_pr_state` would not close.
- Extends the designate-before-close ordering (GH-4826) to the review-feedback path: new `spawnReviewIssue` seam sets `TerminalLabel` the instant `CreateReviewIssue` returns a live issue, strictly before `handleReviewRequested` closes the PR. `CreateReviewIssue` now records a durable spawned-fix claim the same way `CreateFailureIssue` does.

## Design choice (per issue's Refs, item 1)

Chose the durable-claim fallback over a `terminal_label` DB column: the crash happens *before* `persistPRState` runs regardless of what columns that call would write, so a new column wouldn't close the window. The `autopilot_spawned_fixes` table is already written synchronously ahead of the PR close, so consulting it in `notifyExternalClose` closes the gap with no migration.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/autopilot/... -run 'GH4841|GH4826' -v` — new crash/rehydration tests pass for both the pre-merge CI-failure rung and the review-requested rung (seed PR state → invoke close handler directly, simulating a crash with no persist → `RestoreState` on a fresh controller sharing the store → external-close scan asserts `pilot-failed` is applied, not `pilot-retry-ready`)
- [x] `go test ./internal/autopilot/...` — full package suite passes, including existing GH-4826 tests
- [x] `go vet ./internal/autopilot/...`

### DIFF
--- a/.agent/tasks/gh-4841.md
+++ b/.agent/tasks/gh-4841.md
@@ -1,0 +1,43 @@
+# GH-4841
+
+**Created:** 2026-08-11
+
+## Problem
+
+GitHub Issue GH-4841: fix(autopilot): recovery-owner designation must survive restart — persist it and extend the seam to the review path
+
+# fix(autopilot): recovery-owner designation must survive restart — persist it and extend the seam to the review path
+
+**Status**: 📋 Ready to dispatch
+**Created**: 2026-08-11
+**Assignee**: Pilot
+
+## Context
+
+PR#4840 (GH-4826) built the right seam: `spawnFailureIssue` (internal/autopilot/controller.go:2462) designates exactly one recovery owner per CI-failure close, and TerminalLabel is set iff `CreateFailureIssue` returns a live issue. Post-merge review found the designation is **in-memory only** — one daemon restart in the wrong window resurrects the exact double-arm incident the PR fixed. The daemon has a documented history of mid-flow restarts (GH-4807).
+
+Crash window: `TerminalLabel` (internal/autopilot/types.go:1169) has **no column** in `SavePRState`/`GetPRState` (internal/autopilot/state_store.go:638-727). A crash between the PR close (controller.go:2720) and the end-of-ProcessPR persist (controller.go:2093) rehydrates the PR label-less at StageCIFailed; `checkExternalMergeOrClose` (processAllPRs, controller.go:6881) runs before any handler, has no StageCIFailed guard, and the self-close markers (controller.go:582-591) are also in-memory — so `notifyExternalClose` (controller.go:7455-7458) defaults to `pilot-retry-ready` while the spawned fix issue is live. Double-arm, restored by restart.
+
+Same shape on the review path: `handleReviewRequested` still closes before designating — direct `CreateReviewIssue` (controller.go:3007), close (controller.go:3050), designation after (controller.go:3065), in-memory.
+
+## Implementation
+
+1. **Persist the ownership decision.** Either add a `terminal_label` column to the PR-state store (SavePRState/GetPRState/RestoreState roundtrip, migration included) **or** make `notifyExternalClose` consult the durable `autopilot_spawned_fixes` claim before defaulting to retry-ready. Prefer whichever is smaller and closes BOTH the pre-merge (controller.go:2672) and post-merge (controller.go:4283) rungs; state your choice in the PR description.
+2. **Extend the designate-before-close seam to `handleReviewRequested`** — route it through `spawnFailureIssue`-style ordering (designate, then close), or persist its designation the same way as (1).
+3. **Crash/rehydration test**: simulate close → crash (no persist) → RestoreState → external-close scan; assert retry is NOT armed while a spawned fix issue is designated. This is the test PR#4840's review flagged as missing (structural string-match only today, gh4826_test.go:443).
+
+Out of scope: owner-death recovery (preflight-rejected fix issues) — separate issue; iteration-cap sites (controller.go:2594/2616) spawn nothing, leave them.
+
+## Acceptance
+
+- Daemon restart at ANY point between PR close and persist does not arm `pilot-retry-ready` when a spawned fix issue is the designated owner (proven by the rehydration test, not by inspection).
+- Review path shares the designate-before-close ordering or durable designation.
+- Existing GH-4826 tests still pass; no schema change beyond the one migration if option (1a) chosen.
+
+## Refs
+
+- Review verdict: https://github.com/qf-studio/pilot/pull/4840#issuecomment-5253781955 (D1, D4)
+- Prior work: PR#4840 (GH-4826), incident GH-4820, restart history GH-4807
+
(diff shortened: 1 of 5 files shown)


alekspetrov:
## Pre-merge review — PR#4846 vs issue #4841 / TASK-467

**Verdict: HOLDS WITH DEFECTS** — spec items 1–3 are all delivered and the rehydration tests are genuine (mutation-verified), but the durable-claim fallback has no owner-health check, which is exactly the risk pre-recorded by the #4849 review wave. D1 needs a follow-up issue filed before/at merge; it does not block the primary acceptance criterion.

### Defects

**D1 (major) — fallback designates a dead owner: no health check on the durable claim.**
`internal/autopilot/controller.go` notifyExternalClose fallback (new hunk, merged-tree ~7491) + `internal/autopilot/state_store.go` `HasSpawnedFixForPR`. The fallback trusts `issue_number > 0` from `autopilot_spawned_fixes` and never fetches the fix issue. Two failure scenarios:

- *No race needed:* fix issue closed by a human during daemon downtime → restart → external-close scan labels the source `pilot-failed` with "fix issue #N already owns this work" while #N is a corpse. No recovery ever fires: #4849's reactions are event-driven only (preflight-decline hook, dedup-path re-check inside `CreateFailureIssue`) and neither can trigger again for a closed, untracked PR. **Permanent strand.** Pre-4846, the identical timeline self-healed: no fallback → `pilot-retry-ready` → retry proceeds.
- *Interaction with #4849's D1 ordering race (risk b):* restart inside the close→persist window; after restart the SDK poller preflight-declines the fresh fix issue **before** the controller's external-close tick lands `pilot-failed`. `ReactToDeclinedFixIssue` → `reactToDeadFixIssue` skips at `owner_death.go:144-149` ("source issue not currently designated" — source has no `pilot-failed` yet), consuming the one-shot reaction. The fallback then lands `pilot-failed` pointing at the declined owner. So 4846 doesn't just widen the D1 window by (downtime + restore + poll tick) — it converts the race's crash-path outcome from benign (retry-ready arrives anyway) to permanent strand.

Fix is small on the merged tree: after `fixIssue > 0`, `GetIssue` + `classifyOwnerHealth` (both exist post-#4849); `ownerDead` → fall through to retry-ready + `emitOwnerDeathAlert`; API error → trust the claim (mirrors GH-4842's dedup-path fail-open). Note: the PR branch is based at `fef6e714`, which predates `owner_death.go`, so the author could not have called the helper — that explains the gap, it doesn't close it.

**D2 (minor) — branch base is 7 commits behind main (predates #4849/#4850).**
Verified locally: `git merge-tree` preview vs current main is conflict-free, builds, and the **full** `./internal/autopilot` suite passes on the merged tree, including all GH-4826, GH-4841, and owner-death tests. No action if merged soon; re-verify if main moves.

**D3 (minor, pre-existing exposure) — review-path claim is recorded after issue creation.**
`internal/autopilot/feedback_loop.go` (CreateReviewIssue hunk): crash between `CreatePilotIssue` and `ClaimSpawnedFix` leaves an orphan review issue; on restart the PR is still open + changes-requested, `handleReviewRequested` re-runs and mints a duplicate revision issue (no dedup gate, by design per the hunk comment). Not a regression — review issues had zero dedup before this PR — but the new claim machinery could have closed it. Fold into the D1 follow-up or accept.

**D4 (nit) — `HasSpawnedFixForPR` ordering.**
`ORDER BY created_at DESC` on a second-resolution DATETIME with no rowid tiebreaker; two claims in the same second pick arbitrarily, and `RecordSpawnedFixIssue` re-records (repeat review rounds) don't bump `created_at`. Harmless today.

### Spec conformance (verified, not asserted)

1. **Persistence choice:** option (1b) — durable-claim consult in `notifyExternalClose` — implemented and stated in the PR description with correct reasoning (a `terminal_label` column can't close a window that ends before `persistPRState` runs; the claim row is written synchronously *before* the close). No schema change — allowed under acceptance. Pre-merge rung proven by test; post-merge rung closed by construction (same `CreateFailureIssue` claim, same shared `notifyExternalClose`), untested directly but residual exposure is ~0: the `Merged=true` path never reaches `notifyExternalClose`, and the source issue is normally already closed at that rung (controller.go:~4290 comment).
2. **Review path:** gets *both* halves — designate-before-close via the new `spawnReviewIssue` seam (TerminalLabel set the moment `CreateReviewIssue` returns, strictly before `ClosePullRequest`), and a durable claim recorded inside `CreateReviewIssue` before the handler can close. The old designate-after-close at the tail of `handleReviewRequested` is correctly removed.
3. **Crash/rehydration tests are real behavior, not structure:** `gh4841_test.go` seeds a persisted row, calls the close handler directly (so end-of-cycle persist never runs), restores into a *fresh controller sharing the same SQLite store*, asserts `TerminalLabel` is empty post-restore (anti-vacuity), then drives `checkExternalMergeOrClose` and asserts labels from decoded httptest request bodies. **Mutation check:** disabling the fallback (`else if false && ...`) fails both new tests — they execute the claimed path. No argument-discarding mocks.
4. **GH-4826 tests:** all 4 pass on the merged tree.
5. **Drive-bys:** none. Five files, all in scope (`.agent/tasks/gh-4841.md` is the standard Pilot scratch doc).

### Notes / nits

- Iteration-cap `TerminalLabel` sites (controller.go:2619, :3002) remain crash-lossy — a restart in that window arms retry past the cap. Explicitly out of scope per the issue; noting so it isn't mistaken for covered.
- Prefix match `fix:pr<N>:%` is safe against PR-number prefix collisions (the `:` after the number terminates it) and against LIKE-injection (pattern built from an int only).
- Repo scoping preserved: fallback keys on `c.repoKey()` = the same `owner/repo` the claims are written under (GH-3903 discipline).
- `.agent/tasks/gh-4841.md` duplicates TASK-467 content — archive on merge per doc-maintenance rules.

### Bottom line

The mechanism is the right one and the tests prove the headline acceptance criterion: restart anywhere between close and persist no longer arms `pilot-retry-ready` while a **live** fix issue owns the work. The dead-owner sub-case (risk a) regresses from self-healing to permanent strand, and the rehydration delay sharpens #4849's decline-race (risk b) into that same strand. File the follow-up for the owner-health check in the fallback (one small delta now that `classifyOwnerHealth` is on main); with that tracked, merge is acceptable.

Issue #4843, opened 2026-08-11
fix(gateway): chat API — gateway-mode wiring hole + latestSeq for reset detection


alekspetrov:
# fix(gateway): chat API — gateway-mode wiring hole + latestSeq for reset detection

**Status**: 📋 Ready to dispatch
**Created**: 2026-08-11
**Assignee**: Pilot

## Context

PR#4838 (GH-4835) shipped the C17 operator chat API (poll-drain web transport for the shared comms brain). Post-merge review confirmed the contract but found two defects and two riders. The console leg (pilot-console#115) anchors on this endpoint — D2 blocks it from implementing the documented reset rule.

- **D1 (medium-high)** — chat silently absent in gateway mode when no polling adapter is enabled: `needsPollingInfra` (cmd/pilot/main.go:663) omits chat, so `gwRunner` is nil → `WithChatHandler(nil, …)` (cmd/pilot/main.go:1072) → guard at internal/pilot/pilot.go:815 skips `SetChatAPI` → routes 404 while cmd/pilot/main.go:1073 logs "Chat API enabled in gateway mode". A webhooks-only gateway daemon with the console chat panel hits this.
- **D2 (medium)** — reset detection unimplementable: the documented rule says a client seeing `after` > latest seq must re-poll from 0, but the GET events response never exposes the latest seq — internal/adapters/web/api.go:124 discards it. After daemon restart or 1h conversation expiry, stale-cursor clients poll into silence, then get partial replay with no signal. Same gap hides drop-oldest truncation past the 500-event cap.
- **D3 (low)** — `events` serializes as `null` (not `[]`) for unknown/expired conversations (internal/gateway/messenger.go:241 returns nil slice).
- **D4 (low)** — `TestAPI_Dispatch_UsesGivenContextNotCancelledOne` (internal/adapters/web/api_test.go:429) never cancels a context; the spec's cancellation-survival behavior is unasserted.

## Implementation

1. **D1**: make chat count toward runner construction — include chat in `needsPollingInfra` (or build the chat runner independently of polling infra); a gateway-mode daemon with chat enabled and zero polling adapters must serve both chat routes. Kill the false "enabled" log on any path that doesn't wire the handler.
[...]

Issue #962, opened 2026-02-13
Add crash recovery and orphan cleanup


alekspetrov:
Parent: GH-951

Implement startup scan for orphan `/tmp/pilot-worktree-*` directories
Current cleanup relies on deferred functions. If Pilot crashes mid-execution, worktrees remain orphaned. Need:
- Startup scan that prunes stale worktrees
- Check for worktrees not in active tracking map
- `git worktree prune` on startup

alekspetrov:
✅ Completed as part of GH-951
PR: Warning: 2 uncommitted changes
https://github.com/alekspetrov/pilot/pull/967

Issue #635, opened 2026-02-09
Wire team project mapping into task scoping


alekspetrov:
## Context

`ProjectAccess` struct and service methods exist in `internal/teams/`:
- `SetProjectAccess(teamID, actorID, projectPath, defaultRole)` 
- `RemoveProjectAccess(teamID, actorID, projectPath)`
- `ListProjectAccess(teamID)` — returns all project mappings
- `CheckProjectAccess(memberID, projectPath, requiredPerm)` — validates member can access project

But tasks execute globally — project access is never validated against team mappings.

**Part of GH-632: Wire remaining 4 ⚠️ features (3/4)**
**Depends on: GH-633, GH-634**

### 1. Validate project scope in GitHub polling
In `cmd/pilot/main.go` GitHub polling loop, when a new issue is picked up:
- If `--team` is set, validate the repo/project is in the team's allowed projects
- Skip issues from repos not mapped to the team (log warning, don't execute)
- This prevents cross-project execution at the poller level (before Runner)

### 2. Add project validation to Telegram handler
In Telegram task handling:
- If team is configured, check project access before dispatching
- Return user-friendly message: "Project X is not assigned to your team"

### 3. CLI project-team management integration
Ensure `pilot project` CLI can show team mappings:
- `pilot project show <name>` should display which teams have access (if teams DB exists)
- `pilot team show <id>` already shows projects — verify it works end-to-end

### 4. Set MemberID from project default role
When team is configured but specific member is unknown (e.g., webhook-triggered):
- Use project's `DefaultRole` to determine access level
- Fall back to team-level default if no project-specific role

## Tests
- Test: GitHub poller skips issues from unmapped repos
- Test: Telegram rejects tasks for unmapped projects  
- Test: `pilot project show` displays team access info

## Files
- `cmd/pilot/main.go` — project validation in polling loop
- `internal/adapters/telegram/handler.go` — project access check before dispatch
- `cmd/pilot/project.go` — show team mappings in project info

## Acceptance
- `make test` passes
- Issues from unmapped repos are skipped with warning log
- Team-scoped execution only runs on assigned projects

alekspetrov:
⚠️ Pilot execution completed but no changes were made.

**Duration:** 32m7.889s
**Branch:** `pilot/GH-635`

No commits or PR were created. The task may need clarification or manual intervention.

PR #4848, opened 2026-08-11
fix(gateway): chat API gateway-mode wiring hole + latestSeq for reset detection

alekspetrov:
## Summary

Post-merge review follow-up on PR#4838 (GH-4835) found four defects in the poll-drain chat transport. Fixes D1-D4:

- **D1** — gateway mode silently dropped chat routes (404) when no polling adapter was configured, because `needsPollingInfra` never considered chat, leaving `gwRunner` nil. Adds `gatewayChatNeedsOwnRunner` to build a chat-scoped runner independently of `needsPollingInfra`, and turns the misleading "Chat API enabled" log into a warning on any path that still leaves the handler unwired.
- **D2** — GET events never exposed `latestSeq`, making the documented stale-cursor reset rule unimplementable by clients. Threads `latestSeq` through `WebMessenger` → `API.Events` → `ChatAPI` → `chatEventsResponse`, with the reset/truncation contract documented at each hop.
- **D3** — unknown/expired conversations serialized `events` as `null` instead of `[]`. Fixed at the source in `WebMessenger.Events`.
- **D4** — `TestAPI_Dispatch_UsesGivenContextNotCancelledOne` never actually cancelled a context, leaving the cancellation-survival behavior unasserted. Fixed to actually cancel and assert dispatch completes.

Out of scope (per issue): streaming/SSE, auth changes, comms-brain behavior, console-side work (pilot-console#115).

## Test plan

- [x] `go build ./...`
- [x] `go test ./cmd/pilot/... ./internal/adapters/web/... ./internal/gateway/... ./internal/pilot/...` (including `-race`, repeated runs)
- [x] New integration test (`internal/pilot/chat_gateway_wiring_test.go`): gateway mode, chat enabled, zero polling adapters → POST 202 + GET drains events (was 404 pre-fix)
- [x] `latestSeq` reset-detection contract covered in `internal/gateway/chat_test.go` / `internal/adapters/web/api_test.go`
- [x] Unknown conversation returns `{"events": []}`

### DIFF
--- a/cmd/pilot/main.go
+++ b/cmd/pilot/main.go
@@ -382,6 +382,28 @@ func main() {
 	}
 }
 
+// gatewayChatNeedsOwnRunner reports whether gateway mode must build a
+// dedicated executor.Runner for the chat API on its own, separate from the
+// needsPollingInfra block. GH-4843 (D1): chat is an HTTP endpoint, not a
+// polling adapter — before this fix, chatEnabled didn't factor into
+// needsPollingInfra at all, so a webhooks-only gateway daemon (chat enabled,
+// zero polling adapters) left gwRunner nil, silently wiring
+// WithChatHandler(nil, ...) and leaving the chat routes unregistered (a 404)
+// despite the startup log claiming "Chat API enabled in gateway mode". When
+// needsPollingInfra is already true, that block builds gwRunner itself, so
+// this only needs to cover the gap case.
+//
+// Decision: fixed by building a second, chat-scoped runner rather than
+// folding chatEnabled into needsPollingInfra's own condition. The latter
+// would also pull in that block's polling-only side effects (memory store +
+// dispatcher, teams RBAC, approval handlers, autopilot controller, alerts
+// engine) for a deployment that only asked for the chat HTTP endpoint —
+// scope creep with real side effects (e.g. autopilot starting to act on PRs)
+// for config that never opted into it.
+func gatewayChatNeedsOwnRunner(chatEnabled, needsPollingInfra bool) bool {
+	return chatEnabled && !needsPollingInfra
+}
+
 func newStartCmd() *cobra.Command {
 	var (
 		dashboardMode bool
@@ -665,6 +687,14 @@ Examples:
 					cfg.Adapters.GitHub.Polling != nil && cfg.Adapters.GitHub.Polling.Enabled) ||
 				(slackFlagSet && hasSlack) ||
 				adapterPollerEnabled
+			// GH-4843 (D1): chat is an HTTP endpoint, not a polling adapter, so
+			// it must NOT gate on needsPollingInfra the way Telegram/Slack/GitHub
+			// polling do — a webhooks-only gateway daemon with chat enabled and
+			// zero polling adapters still needs a runner to dispatch chat tasks.
+			// Checked separately from needsPollingInfra below so a chat-only
+			// deployment doesn't also spin up autopilot/approvals/alerts infra
+			// it never asked for.
+			chatEnabled := cfg.Adapters.Chat != nil && cfg.Adapters.Chat.Enabled
 
 			// Shared infrastructure for polling adapters
 			var gwRunner *executor.Runner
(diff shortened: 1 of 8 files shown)


alekspetrov:
## Post-merge review — GH-4843 (D1–D4)

**Verdict: APPROVED post-merge. All four defects fixed, acceptance met, no regressions found. No follow-up issue required; three low/nit observations below.**

Verified at merge commit 3424cdaf: `go build ./...` clean; `internal/adapters/web`, `internal/gateway`, `internal/pilot` green under `-race -count=1`; `cmd/pilot` decision test green.

### Per-item

- **D1 ✓** — `gatewayChatNeedsOwnRunner` (cmd/pilot/main.go:397) chained as `else if` off the `needsPollingInfra` block (main.go:1071), which itself unconditionally builds `gwRunner` or errors out — so at the wiring point (main.go:1120) `gwRunner` is non-nil on every chat-enabled path. The false "Chat API enabled" log is gone: Info only when the runner exists, Warn on the (now-defensive) unwired path (main.go:1122-1125). Building a narrow chat-scoped runner instead of folding chat into `needsPollingInfra` is the right call — folding in would have dragged autopilot/approvals/alerts side effects into chat-only deployments. Polling-mode chat wiring (main.go:2667-2682) untouched; when `needsPollingInfra` is true the path is byte-identical to before, so polling mode is unaffected.
- **D2 ✓** — `latestSeq` on `chatEventsResponse` (internal/gateway/chat.go:339), threaded `WebMessenger.Events` → `API.Events` (internal/adapters/web/api.go no longer discards it) → handler. Value semantics verified against messenger internals: `appendLocked` increments-then-assigns, so `c.nextSeq` holds the newest *assigned* seq and `latestSeq` is correct (0 for no buffer), consistent with the 202 accept-time seq. Reset + truncation contract documented at all three hops (api.go Events doc, chat.go response doc, messenger.go Event doc).
- **D3 ✓** — fixed at the source (`[]Event{}` for unknown/expired, internal/adapters/web/messenger.go:241) with the known-conversation path already non-nil via `make`, plus a belt-and-braces nil→`[]` conversion in the handler (chat.go:176-178) covering any other `ChatAPI` impl. Wire-level test asserts the raw bytes contain `"events":[]`.
- **D4 ✓** — the test now creates and genuinely cancels a context, asserts `requestCtx.Err() != nil` in setup, and asserts dispatch completes (internal/adapters/web/api_test.go:98-133). See observation 1 on discriminating power.

### Acceptance

- Gateway-mode-no-pollers integration test: `TestGatewayMode_ChatEnabledNoPollingAdapters_RoutesServe` (internal/pilot/chat_gateway_wiring_test.go) — real HTTP listener, chat enabled, zero polling adapters, POST 202 + GET drains a real event. Passes under `-race`. It asserts what it claims for the callee side; the caller-side decision is covered separately by `TestGatewayChatNeedsOwnRunner` (cmd/pilot/wiring_test.go:749). Nothing executes `newStartCmd` end to end, but the two halves plus the reviewed `else if` placement close the chain.
- Stale-cursor reset detection: `TestAPI_Events_LatestSeqDetectsReset` (api_test.go:282) and `TestHandleChatEvents_LatestSeqSignalsReset` (chat_test.go:435) both simulate a fresh post-restart buffer with cursor > latestSeq and prove reset is detectable from the response body alone. They assert what they claim.
- Unknown conversation → `{"events": []}`: covered at bytes level (chat_test.go:412) and API level (api_test.go:252).

### Observations (non-blocking)

1. **Low — D4 test is documentation more than enforcement** (internal/adapters/web/api_test.go:98-133). The cancelled `requestCtx` is never passed into the code under test, so cancelling it cannot influence `Dispatch` — the test would pass even if `Dispatch` mishandled contexts. The load-bearing assertion for this property is the ctx-identity check in the handler test (`gotCtx == s.daemonCtx`, internal/gateway/chat_test.go:184) plus `Dispatch` passing `dispatchCtx` verbatim to the goroutine (api.go:110). Combined coverage is adequate; fine to leave as-is.
2. **Low — chat-only runner skips two setup steps the polling path runs** (cmd/pilot/main.go:1071-1086): `wireProjectAccessChecker` (GH-635 team project access on the runner) and the GH-962 orphaned-worktree startup sweep. A chat-only deployment with `teams` configured runs chat-dispatched tasks without the project-access checker, and with `executor.use_worktree` on, crashed executions' worktrees are never swept. Both plausibly intentional per the documented scope-narrowing decision — worth a line in the config docs if chat-only + teams becomes a real deployment shape.
3. **Nit, pre-existing — fixed test port 19097 is used by two packages** that can run concurrently under `go test ./...`: internal/gateway/chat_test.go:392 and internal/config/gateway_auth_test.go:80. Same flake class this PR's second commit fixed for 19199/19099 (the new 19099 is unique tree-wide — verified). Not introduced here; flagging while in the area.
4. **Note** — seq-only reset detection has an inherent blind spot: if a post-restart conversation accumulates events past a stale client's cursor before it polls, the reset is invisible (no epoch/instance ID). This matches what the spec asked for and the 202-seq hint mitigates in practice; recording it so it isn't re-reported as a defect later.

### Console contract (pilot-console#115 §3 / PR#117)

Shape matches the promise: top-level `latestSeq` (int64) on the GET events response beside `conversationId`/`events`, exactly the field name console#115 §2/§3 anchored on. Console#117's relay-verbatim proxy passes it through unchanged. The `events: []` fix also retires the null-relay caveat #115 had to pin — console's null-case test remains valid defensively but the daemon no longer emits null.

Issue #4820, opened 2026-08-09
fix(ci): resolve CI failure from PR #4818


alekspetrov:
# Autopilot: Auto-Generated Fix Request

## Context

- **Original PR**: #4818
- **Original Issue**: #4817
- **Failure Type**: ci_pre_merge
- **SHA**: 6075530
- **Branch**: pilot/GH-4817

## Failed Checks

- [ ] lint

<details><summary>CI Error Logs</summary>

```
=== lint — failing step: golangci-lint ===
2026-08-09T07:32:57.1815612Z Cache restored successfully
2026-08-09T07:32:57.1916167Z Cache restored from key: setup-go-Linux-x64-ubuntu24-go-1.25.12-ba73b9216bf293414b54458d88fc515eff5a5b2333fefc0d27ebf9b7ddf2a747
2026-08-09T07:32:57.1938060Z go version go1.25.12 linux/amd64
2026-08-09T07:32:57.1938260Z 
2026-08-09T07:32:57.1938583Z ##[group]go env
2026-08-09T07:32:57.2609514Z AR='ar'
2026-08-09T07:32:57.2609919Z CC='gcc'
2026-08-09T07:32:57.2610244Z CGO_CFLAGS='-O2 -g'
2026-08-09T07:32:57.2610636Z CGO_CPPFLAGS=''
2026-08-09T07:32:57.2611026Z CGO_CXXFLAGS='-O2 -g'
2026-08-09T07:32:57.2611453Z CGO_ENABLED='1'
2026-08-09T07:32:57.2611836Z CGO_FFLAGS='-O2 -g'
2026-08-09T07:32:57.2612240Z CGO_LDFLAGS='-O2 -g'
2026-08-09T07:32:57.2612604Z CXX='g++'
2026-08-09T07:32:57.2612907Z GCCGO='gccgo'
2026-08-09T07:32:57.2613229Z GO111MODULE=''
2026-08-09T07:32:57.2613868Z GOAMD64='v1'
2026-08-09T07:32:57.2614216Z GOARCH='amd64'
2026-08-09T07:32:57.2614570Z GOAUTH='netrc'
2026-08-09T07:32:57.2614930Z GOBIN=''
2026-08-09T07:32:57.2615297Z GOCACHE='/home/runner/.cache/go-build'
2026-08-09T07:32:57.2615767Z GOCACHEPROG=''
2026-08-09T07:32:57.2616112Z GODEBUG=''
[...]

Issue #4309, opened 2026-07-14
fix(autopilot): idempotency guard on CI-fix issue creation — dedup in shared store (kills duplicate fix-issue spawning)


alekspetrov:
## Context

Autopilot filed 4 identical CI-fix issues (#4301/#4302/#4304/#4305) for one PR+failure. Root cause (PRIMARY, per `.agent/system/incident-duplicate-cifix-2026-07-14.md`): `FeedbackLoop.CreateFailureIssue` (`internal/autopilot/feedback_loop.go:137-189`) does ZERO existence checking — it builds title (`:198`) + body (`:214`) and unconditionally calls `ghissue.CreatePilotIssue`. No search for an existing open fix-issue, no persisted dedup key. The only respawn suppression is per-process in-memory `activePRs`/`removePR` — useless across ticks, release re-scans, and concurrent daemons. This is the 3rd occurrence (prior: #4258; #4261-63).

## Implementation

Add a durable dedup claim in the SHARED SQLite store, checked before every fix-issue creation:
1. `internal/autopilot/state_store.go`: new `spawned_fix_issues` table (PRIMARY KEY `(repo, dedup_key)`, columns: created_at, issue_number, failure_type). Methods `RecordSpawnedFix(repo, key, issueNum)` (INSERT OR IGNORE, returns whether newly-inserted) + `SpawnedFixExists(repo, key)`.
2. `internal/autopilot/feedback_loop.go:137` (`CreateFailureIssue`): compute `dedupKey = fmt.Sprintf("pr%d:%s:%s", prNumber, failureType, sortedFailedChecksSig)` (include sorted failed-check names so a genuinely NEW failure spawns but same-check repeats are blocked); check `SpawnedFixExists` → skip+log if present; else create, then `RecordSpawnedFix`.
3. Wire the guard at both call sites: `controller.go:2621` (post-merge) and `controller.go:1589` (pre-merge/CI-fix).
4. Belt-and-suspenders: before creating, GitHub-search open issues by exact title + pilot label (covers a lost DB row / fresh clone).

## Acceptance

- [ ] Two sequential `CreateFailureIssue` calls for the same (repo, pr, failureType, checks) create exactly ONE issue; the second is a logged no-op
- [ ] A different failed-check signature on the same PR DOES create a new issue
- [ ] Guard uses the shared store so two processes hitting it concurrently still yield one issue (INSERT OR IGNORE race-safe)
[...]

PR #4849, opened 2026-08-11
fix(autopilot): owner-death recovery — re-arm or escalate when a designated fix issue dies

### LINKED ISSUE #4842: fix(autopilot): owner-death recovery — a designated fix issue that dies must re-arm the source or escalate

alekspetrov:
## Summary

After PR#4840 (GH-4826), a CI-failure close designates exactly one recovery owner (the spawned fix issue) via `prState.TerminalLabel`. If that owner itself dies — declined at preflight before dispatch, or closed without ever shipping a merged PR — nothing re-armed the source or escalated it, stranding the source permanently as `pilot-failed` with zero live owners.

Two reactive seams cover both death modes without a new poller:

- `storeExecutionSaver.SaveDeclinedExecution` (`cmd/pilot/main.go`) is called synchronously by the SDK poller on every preflight decline; it now reacts via `Controller.ReactToDeclinedFixIssue` when the declined task is a Pilot-spawned fix issue.
- `FeedbackLoop.CreateFailureIssue`'s dedup path (`feedback_loop.go`) now checks the previously-designated fix issue's open/shipped/dead state before handing it back out, and falls through to mint a replacement (overwriting the dead issue's dedup row) instead of re-designating a closed-unmerged owner.

Both paths trace the dead fix issue back to its source via the existing `Depends on: #N` body convention (no new persistence, per scope) and either re-arm the source with `pilot-retry-ready` or escalate to `pilot-needs-human` when retries are exhausted, firing an alert through the existing alerts engine either way.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/autopilot/... ./cmd/pilot/...`
- [x] Table-driven matrix: preflight-declined → rearm/escalate, source-already-closed → no-op, dedup-returns-open/shipped (not death) vs dedup-returns-closed-unmerged (death → replacement minted, dedup row overwritten, exactly one alert)

Closes GH-4842.

### DIFF
--- a/cmd/pilot/main.go
+++ b/cmd/pilot/main.go
@@ -4146,11 +4146,22 @@ func (c terminalCompletionChecker) InvalidateCompletion(taskID, projectPath stri
 // GH-2802: Persists pre-flight rejection records for observability.
 type storeExecutionSaver struct {
 	store *memory.Store
+	// controller is optional (nil for a repo with no autopilot controller
+	// wired) — when set, a preflight decline of a Pilot-spawned fix issue
+	// reacts via the owner-death path (GH-4842) instead of only being
+	// recorded for observability.
+	controller *autopilot.Controller
 }
+
+// ownerDeathReactTimeout bounds the synchronous owner-death reaction
+// (GitHub issue fetch + label/comment writes) fired from inside
+// SaveDeclinedExecution, which the SDK poller calls inline on every
+// pre-flight decline.
+const ownerDeathReactTimeout = 15 * time.Second
 
 func (s storeExecutionSaver) SaveDeclinedExecution(taskID, projectPath, status, reason string) error {
 	now := time.Now()
-	return s.store.SaveExecution(&memory.Execution{
+	err := s.store.SaveExecution(&memory.Execution{
 		ID:          fmt.Sprintf("%s-preflight-%d", taskID, now.UnixNano()),
 		TaskID:      taskID,
 		ProjectPath: projectPath,
@@ -4159,6 +4170,20 @@ func (s storeExecutionSaver) SaveDeclinedExecution(taskID, projectPath, status, 
 		CreatedAt:   now,
 		CompletedAt: &now,
 	})
+
+	// GH-4842: a preflight decline is owner death for a Pilot-spawned fix
+	// issue — react (re-arm/escalate its source) using the same signal the
+	// SDK already produces here, instead of adding a new poller.
+	if status == "declined-preflight" && s.controller != nil {
+		var issueNum int
+		if _, scanErr := fmt.Sscanf(taskID, "GH-%d", &issueNum); scanErr == nil && issueNum > 0 {
+			ctx, cancel := context.WithTimeout(context.Background(), ownerDeathReactTimeout)
+			s.controller.ReactToDeclinedFixIssue(ctx, issueNum, reason)
+			cancel()
+		}
+	}
+
+	return err
 }
 
 // warnIfMetricsScopeEmpty logs a startup warning when a configured
--- a/cmd/pilot/poller_github.go
+++ b/cmd/pilot/poller_github.go
@@ -481,7 +481,10 @@ func startGithubSDKPollerForRepo(ctx context.Context, deps *PollerDeps, log *slo
 				repoFullName: target.repoFullName,
 			}
 			if deps.Store != nil {
-				pollerDeps.ExecutionSaver = storeExecutionSaver{store: deps.Store}
+				// GH-4842: wire the per-repo controller (may be nil) so a
+				// preflight decline of a Pilot-spawned fix issue can react
+				// via the owner-death path.
+				pollerDeps.ExecutionSaver = storeExecutionSaver{store: deps.Store, controller: controller}
 			}
 		}
 	}
--- a/internal/autopilot/controller.go
+++ b/internal/autopilot/controller.go
@@ -1839,6 +1839,9 @@ func (c *Controller) SetOnIssueDone(fn func(issueNumber int)) {
 // SetOnIssueDone above. GH-3954.
 func (c *Controller) SetAlertsEngine(engine alertSink) {
 	c.alertsEngine = engine
+	if c.feedbackLoop != nil {
+		c.feedbackLoop.SetAlertsEngine(engine)
+	}
 }
 
 // OnPRCreated registers a new PR for autopilot processing.
--- a/internal/autopilot/feedback_loop.go
+++ b/internal/autopilot/feedback_loop.go
@@ -161,6 +161,7 @@ type FeedbackLoop struct {
 	issueLabels  []string
 	learningLoop *memory.LearningLoop // GH-1979: optional, annotates issues with known patterns
 	stateStore   *StateStore          // GH-4307: optional, dedups fix-issue creation across retries/daemons
+	alertsEngine alertSink            // GH-4842: optional, observes owner-death (dead designated fix issue) events
 	log          *slog.Logger
 }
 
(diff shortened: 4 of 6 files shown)


alekspetrov:
## Post-merge review — GH-4842 owner-death recovery

**Verdict: holds. Merge stands — the core mechanism is correct, no new poller was added, and the dedup hole is closed. One medium-severity ordering race and a handful of smaller follow-ups below.**

### Verified

- **Both death modes hooked into existing seams, no new poller/goroutine** — preflight-declined via `storeExecutionSaver.SaveDeclinedExecution` (called inline by the SDK poller, exactly once per decline); closed-unmerged via the dedup owner-health check in `CreateFailureIssue`. `owner_death.go` contains no goroutines or tickers. Spec preference honored.
- **Re-arm respects retry generations** — `rearmDeadOwnerSource` adds only `pilot-retry-ready` and leaves `pilot-retry-1/2` intact; the SDK poller's ladder (poller.go:1346-1380) advances generation from those retained labels. Escalation gates on `pilot-retry-exhausted`, which is exactly the ladder's terminal state. Correct.
- **Dedup can no longer re-designate a closed-unmerged owner** — health check at feedback_loop.go:276-296, replacement minted, dedup row overwritten (test asserts the 555 overwrite + exactly one alert). The GH-4309 open-issue search can't rematch the dead (closed) issue. Fail-open on lookup error is logged.
- **Alerts wiring is construction-order safe** — `feedbackLoop` is built inside `NewController` (controller.go:841), so the `SetAlertsEngine` forwarding (controller.go:1840-1843) always reaches it.
- **Matrix test** — classify/parse/reaction/dedup rows all assert exactly-one-reaction + exactly-one-alert (or none). Body-convention parse is guarded by the `autopilot-meta` marker and line anchoring against the epic-decomposition false positive.

### Defects

**D1 (Medium) — decline-vs-designation ordering race loses the recovery event permanently.** `reactToDeadFixIssue` (owner_death.go:296-301) skips when the source lacks `pilot-failed`. But that label is applied by `notifyExternalClose` *on the next monitor tick after* the close (controller.go:2609-2612 says so explicitly), while the SDK poller can pick up the freshly-minted fix issue and decline it in the same window. Decline-first ordering: guard skips (Info log only, no alert) → the decline never re-fires (`handlePreFlightReject` adds `needs-clarification` and filters the issue on all later polls — never re-judged) → next tick applies `pilot-failed` → source stranded with a permanently sidelined open owner, which the dedup health check classifies `ownerAlive`. This is the exact GH-4842 bug shape surviving in one tick-ordering. The test row "source not currently designated — no reaction" enshrines the losing ordering as correct. Fix: gate on the durable spawned-fix claim (written before the close in *all* orderings; trivial via `HasSpawnedFixForPR` once #4846 lands) or on controller `prState.TerminalLabel`, instead of the eventually-consistent GitHub label.

**D2 (Medium-Low) — closed-unmerged detection is opportunistic only.** Death mode (b) is detected solely when the same failure signal re-ticks through `CreateFailureIssue` (post-merge scan re-discovery, second daemon). A designated owner closed by a human strands the source with no alert until an identical signal recurs — for the pre-merge path (PR already closed) that is never. "Alerts fire on every owner-death event" holds only opportunistically. Existing no-new-poller seam candidate: the reconciler / external-close scan tick.

**D3 (Low) — `parseFixIssueSource` takes the first `^Depends on: #N$` match, but `generateBody` places the CI log excerpt (feedback_loop.go:427-446) *before* the real Depends-on line (:475).** A column-0 "Depends on: #N" inside a log excerpt (this repo's own test fixtures contain such lines) misdirects the re-arm/escalation to an arbitrary issue. Use the last match, or resolve via the `autopilot-meta pr:N` field cross-checked against the spawned-fix claim.

**D4 (Low) — one-shot reaction, no failure alert.** `ReactToDeclinedFixIssue`/`reactToDeadFixIssue` Warn-and-return on GitHub fetch errors; since the decline signal fires exactly once, a transient 5xx silently loses the recovery event forever. Also the reaction ctx derives from `context.Background()`, detached from daemon shutdown.

**D5 (Low, test gap) — the preflight seam itself is untested.** `ReactToDeclinedFixIssue`, the `GH-%d` taskID parse, and the `declined-preflight` status gate in `SaveDeclinedExecution` (cmd/pilot/main.go:4183-4190) have no coverage; the matrix drives the shared `reactToDeadFixIssue` only.

**D6 (Info) — re-arm leaves the dead owner open.** A declined fix issue keeps `needs-clarification` and stays open; a human later removing that label revives it alongside the re-armed source — the two-live-arms GH-4820 shape. Consider superseding/closing the dead owner on re-arm. Also `ownerShipped` is proxied by the `pilot-done` label, not merged-PR state — a closed-before-labeled window misclassifies a shipped owner as dead (bounded: one spurious replacement).

### Interaction with #4846 (durable designation, unmerged)

- **No dependency, as required**: this PR reads only GitHub labels, the body convention, and dedup rows — it works against in-memory designation as merged and needs nothing from #4846.
- **No textual conflict**: #4846's feedback_loop.go hunk (CreateReviewIssue claim recording, ~:511) and controller.go hunks (notifyExternalClose fallback, spawnReviewIssue) don't overlap this PR's regions. #4846 is based on the pre-#4849 blob (index 0eeb35a6) so it needs a rebase, but contexts don't collide.
- **Two follow-ups for the #4846 review**: (a) its `HasSpawnedFixForPR` fallback re-designates on claim *existence* without health-checking the claimed issue — after this PR it should run `classifyOwnerHealth` (or the owner-death reaction) so crash-rehydration can't designate an owner that died during downtime; (b) its restart rehydration widens D1's window — after restart the poller can decline the spawned fix issue before the external-close scan applies `pilot-failed`, hitting the same guard skip. Both argue for switching D1's guard to the durable claim once #4846 lands.
- #4846's `CreateReviewIssue` claim recording has no dedup-return path, so the owner-health hole does **not** reopen on the review path.

Issue #4825, opened 2026-08-10
fix(autopilot): failure-issue bodies must include the failing lines, not a top-truncated log excerpt


alekspetrov:
## Context

The auto-generated `CreateFailureIssue` body for #4820 (spawned after PR#4818's CI-failure close, 2026-08-09) truncated the CI logs **before** the actual failing line — the errcheck finding in `gh4405_test.go:167` WAS present in the run log (`gh run view --log-failed` surfaces it in one grep), but the excerpt embedded in the issue body cut off above it.

The preflight judge caught it (`reject_vague`) and the issue died as an orphan — defense-in-depth worked — but that's the backstop doing the primary's job. The spawned issue's body is the fix executor's **evidence**: an issue whose logs omit the failing line either burns an execution on re-deriving what the spawner already had, or produces a vague fix. This is the TASK-459 evidence-quality invariant applied to the fix-issue spawn path.

Current excerpt machinery (anchors @ main 2026-08-11):

- `internal/autopilot/ci_failure_excerpt.go` — GH-4779-era excerpter; its own doc comment (:25) says the OLD behavior "concatenated whole job logs and then hard-truncated"; it now budgets with `truncateKeepingTail` (:190).
- `internal/autopilot/feedback_loop.go:67` — `truncateKeepingTail(strings.Join(excerpt, "\n"), maxChars)`; `:344` — `truncateKeepingTail(bundle, maxCIErrorExcerptChars)`; `:459` — 4000-char bound for issue bodies.
- `internal/autopilot/ci_monitor.go:816` — "Logs are truncated to maxLen total characters to keep issues readable."
- Spawn site: `internal/autopilot/controller.go:2637` → `feedbackLoop.CreateFailureIssue`.

## Implementation

1. **Audit first**: reproduce how #4820's body lost the errcheck line despite the GH-4779 excerpter existing — identify which path built that body (which of the truncation sites above ran) and why tail-keeping still dropped the failing line (e.g. failing line mid-log with a long tail of teardown noise after it).
2. **Match-anchored excerpting**: the failure-issue body generator must extract the *specific failing lines* (lint findings, `--- FAIL` test output, compile errors — the grep-able failure markers per check type) and budget the excerpt AROUND those matches (context lines before/after), not from the top or tail of the log. Fall back to current tail-keeping only when no marker matches.
3. Keep the existing total-size bounds (GitHub body limits, `:459`); the change is what the budget is spent on, not how much.
[...]

PR #4850, opened 2026-08-11
fix(autopilot): reconcile sentinel-bundle truncation with excerpt assembly budget

alekspetrov:
## Summary

- The sentinel-bundle path in `generateBody` blind-cut the pre-assembled excerpt bundle to `maxCIErrorExcerptChars` (4000) while `AssembleFailureExcerptsBody` built it against a 12000-char budget (`failedCheckExcerptBudgetChars`) — on a multi-check bundle between those sizes, the smaller cut sliced through the middle, silently dropping whichever check's excerpt landed there (the exact failure mode GH-4825 fixed for the single-blob path, reintroduced one layer up). Now reuses the same 12000 budget so the second cut is a true no-op in the normal case; fixed the stale "defensive no-op" comment.
- Rider 1: un-anchored the `^FAIL\t` marker — GitHub Actions timestamp-prefixes every log line so the anchored form never fired in production; `--- FAIL:` covers the per-test case.
- Rider 2: made `truncateKeepingTail`'s head/tail cuts rune-safe so a multi-byte UTF-8 rune is never split at either cut point.
- Rider 3: deleted the dead `GetFailedCheckLogs` (superseded by `GetFailedCheckExcerpts`, no production callers) and its now-orphaned tests.

## Test plan

- [x] New test `TestFeedbackLoop_CreateFailureIssue_MultiCheckBundle_MiddleFailureSurvives` — asserts a middle check's failing line survives the sentinel-bundle cut, with a built-in sanity check proving the same fixture fails against the pre-fix 4000-char cut
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] Grep-clean: no remaining production callers of the deleted `GetFailedCheckLogs`


alekspetrov:
## Post-merge review — PASS (nits only, no follow-up warranted)

Verified against merge commit e035d7e6 (tip of main at review time). Build, vet, and full `internal/autopilot` suite pass at that commit.

**1. Bundle cap vs assembly budget — correct.** Approach taken: constant reuse. `generateBody` sentinel path now cuts with `truncateKeepingTail(bundle, failedCheckExcerptBudgetChars)` (feedback_loop.go:496), matching the 12000 budget `AssembleFailureExcerptsBody` builds against. The stale "defensive no-op" comment is replaced with an accurate one that even documents the residual `minTailBudget` overshoot case. Empirically confirmed the fix is load-bearing: reverting that one constant to `maxCIErrorExcerptChars` makes the new test fail with "body should contain the middle check's failing line".

**2. Rider 1 — done.** `^FAIL\t` → `FAIL\t` (feedback_loop.go:35), with a correct rationale comment tying it to the Actions timestamp prefix. Extra un-anchored matches only add excerpt windows, so no false-positive downside.

**3. Rider 2 — done, guarantees hold.** Both cut points snap to rune boundaries via `runeSafeHeadLen`/`runeSafeTailStart`; byte-capping is retained (correct direction vs GitHub's 65536-char limit since bytes >= chars in UTF-8). The snaps are shrink-only and the tail budget is recomputed after the head snap, so total stays <= maxChars. Probe-tested every maxChars in 22..199 against a 1-/2-/3-/4-byte-rune mix: cap never exceeded, output always valid UTF-8. The `budget <= 0` branch still appends the suffix past the cap — pre-existing, unreachable from production callers (min real budget is 200).

**4. Rider 3 — done.** `GetFailedCheckLogs` deleted; its 4-subtest block removed; grep-clean of production callers confirmed on the merged tree (remaining hits are comments only).

**5. Test — solid.** `TestFeedbackLoop_CreateFailureIssue_MultiCheckBundle_MiddleFailureSurvives` (feedback_loop_test.go:1586) sizes the fixture in the exact 4000–12000 divergence band, and its in-test sanity check (legacy 4000 cut must drop the line) guards the fixture against drift. Provably fails pre-change (verified, see item 1). 65536 body-limit assertion present in both the existing GH-4825 test (:1570) and the new one (:1652).

### Nits (cosmetic, fold into a future touch of these files)

- `internal/autopilot/ci_monitor.go:1028` — `GetCheckLogs` doc comment still compares present-tense to the deleted `GetFailedCheckLogs` ("Unlike GetFailedCheckLogs, this includes successful/skipped runs too"). The parallel comment in ci_monitor_test.go:1622 was updated to `GetFailedCheckLogsByCheck`; this one was missed.
- `internal/autopilot/ci_monitor_test.go:1685` — same stale present-tense reference ("Unlike GetFailedCheckLogs (single combined string)").
- `internal/autopilot/feedback_loop.go:494` — the overshoot note attributes bundle-over-budget solely to `minTailBudget`; the un-budgeted `\n\n` joiners (2 bytes per check boundary) also contribute. Trivial.
- Rider 2 shipped without a dedicated rune-safety test (spec didn't require one; validated by review probe here).

No defects at fix/blocker severity.

Issue #4646, opened 2026-07-31
fix(autopilot): required_checks naming a check that never posts on a repo → permanent CI pending — detect the mismatch and fail loudly (GH-4643 misdiagnosis class)


alekspetrov:
# fix(autopilot): required_checks naming a check that never posts on a repo → permanent CI pending — detect the mismatch and fail loudly (GH-4643 misdiagnosis class)

**Created**: 2026-07-31 · **Status**: Draft · **Last Updated**: 2026-07-31
**Repo**: `qf-studio/pilot`

## Incident (2026-07-31, operator-resolved)

18 release-train scopes across auth-service and studio-sdk were stuck or
parked; auth-service went **11 days without a release** (v0.67.3, 07-20 →
07-31) while 27 merged PRs accumulated. Root cause was NOT "no post-merge CI"
(GH-4643's diagnosis) and NOT GH-4384 (fix present and working):

- Both repos inherit the **global** `autopilot.required_checks: [test, lint]`
  (no per-project `ci_checks` override).
- auth-service push-to-main runs: `test`, `Migration Dry-Run Check`,
  `Build & Push Docker Image` — **no run named `lint`**, ever.
- studio-sdk push-to-main runs: `lint`, `Check Secret Patterns`,
  `build-test` — **no run named `test`**, ever.
- `checkRequiredChecks` (`internal/autopilot/ci_monitor.go:242`) seeds every
  required name `CIPending`; a name that never posts stays pending forever,
  and a non-empty allowlist always wins over green discovered runs (GH-4307,
  `ci_monitor.go:227`).
- Post-merge: `handlePostMergeCI`'s GH-4643 no-workflow probe
  (`HasAnyCIConfigured`) passes — CI *exists*, the names just mismatch — so
  every carrier rode the full 30m timeout; GH-4643's cap then parked the
  scopes with the misleading message "this repo likely has no post-merge CI
  configured".

`internal/config/config.go:271` already documents this exact class in a
comment. The `pointer` project already carries the working idiom
(per-project `ci_checks.required_checks: []`).

Operator resolution (all on the box, 15:0x–15:32Z): config backup
(`config.yaml.bak-2026-07-31-cichecks`) + `ci_checks: {required_checks: []}`
added to the auth-service and studio-sdk project blocks · 10 parked + 8
superseded-failed scopes marked `done` (today's rolled-up train carried all
their members) · graceful restart. Two ticks later: auto-discovery found the
[...]

Issue #4478, opened 2026-07-20
fix(autopilot): CI watcher still reports pending after all check-runs green (GH-4384 incomplete) — live repro pointer#108


alekspetrov:
Live repro 2026-07-20 on v2.242.0-8-gb3216f27 (which contains the #4408 fix for GH-4384).

## Context

pointer#108 (branch pilot/GH-107, sha f671620):

- 12:18:46Z stage pr_created -> waiting_ci; 12:18:57Z "CI checks discovered" checks=[integration go web]
- All three check-runs completed SUCCESS by **12:20:25Z** (API: integration 12:19:17, web 12:20:07, go 12:20:25)
- autopilot_pr_state still `stage=waiting_ci, ci_status=pending` at ~12:24Z and ~12:26Z — 4–6 min and ~10 poll cycles after green
- PR was then merged externally at 12:25:37Z, so the stall window observed is bounded, but ≥8 consecutive polls failing to see green check-runs is not a timing artifact

## Approach

Trace the pointer controller's CI poll path on this repro: `internal/autopilot/ci_monitor.go` polls check-runs (post-#4408), so find why the poll result never reached `autopilot_pr_state` — candidates: discovery mode=manual list pinning to required-checks that never match, per-repo controller polling the wrong owner/repo/sha after the GH-4472 wiring, success result computed but transition gated elsewhere. Add a regression test in the style of gh4384_ci_watcher_test.go reproducing green-check-runs-but-pending.

## Acceptance

- A PR whose check-runs are all green transitions waiting_ci -> ci_passed within one poll interval, on an Actions-only repo (no legacy commit statuses).
- Regression test covering this repro shape.

## Refs

- #4384, #4408 (prior fix), #4415 (restore ci poll), GH-4472 (per-project wiring, in case of interaction)

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4479 |
| Branch | `pilot/GH-4478` |
| Time to merge | 3m55s |

alekspetrov:
held for scope release schedule

Issue #4384, opened 2026-07-16
fix(autopilot): CI watcher polls legacy combined-status API while discovery uses check-runs — every Actions-only new repo times out at waiting_ci with green checks


alekspetrov:
## Context

**Every PR on a newly-onboarded repo dies at `waiting_ci` with "CI timeout after 30m0s" while its checks are green.** Observed on qf-studio/pointer (onboarded 2026-07-16, first cycle): PRs #5/#6/#7 all followed `pr_created → waiting_ci → "CI checks discovered" (checks=[go] / [web]) → 30m timeout → failed`, ledger `ci_status` frozen at `pending`, `gh pr checks` shows pass on all three (PR #7's check completed ~1 min into the watch — not a stale-SHA race). The Slack pre_merge approval never fires because the pipeline dies first. Same daemon merges qf-studio/pilot PRs normally all day.

**Root-cause evidence (2026-07-16 20:0xZ, PR #7 head `35e110afb568f91805528ebd7e6abedda46a8524`):**

```
GET repos/qf-studio/pointer/commits/{sha}/check-runs → go: status=completed conclusion=success
GET repos/qf-studio/pointer/commits/{sha}/status     → state=pending total_count=0
```

GitHub's legacy combined-status API returns `state=pending` when ZERO commit statuses exist, and GitHub Actions only writes check-runs, never legacy statuses. So: **discovery reads the check-runs API (finds [go]) but the completion watcher polls the legacy combined-status API → phantom "pending" forever.** Established repos don't hit it because non-empty `required_checks` (e.g. pilot's [test, lint]) routes the watch down the check-runs path (see GH-4333→#4342 precedence handling); pointer has no `required_checks`, `ci_checks.mode: auto` — discovery succeeds, watch goes blind.

Ledger rows: `autopilot_pr_state` (5,6,7 × qf-studio/pointer) stage=failed, ci_status=pending, error="CI timeout after 30m0s", created 15:41Z/16:06Z/19:03Z.

Note: an unrelated transient DNS outage (`lookup api.github.com: no such host`, 15:16–15:18Z) appears in the same day's log — it does NOT overlap PR #7's watch window; do not chase it.

Single concern, **do not decompose**. 

## Approach

[...]

Issue #4852, opened 2026-08-11
fix(autopilot): owner-death hardening — health-check the durable-claim fallback, gate decline reaction on the claim (PR#4846/#4849 follow-up)


alekspetrov:
## Context

PR#4846 (GH-4841) made recovery-owner designation durable: `notifyExternalClose` consults the `autopilot_spawned_fixes` claim before defaulting to retry-ready. The pre-merge review (verdict on PR#4846) found the fallback trusts `issue_number > 0` and never fetches the fix issue:

- **Dead owner, no race needed**: fix issue closed by a human during daemon downtime → restart → external-close scan labels the source `pilot-failed` pointing at a corpse. GH-4842's reactions are event-driven only (preflight-decline hook; dedup-path re-check inside `CreateFailureIssue`) and cannot fire for a closed, untracked issue → permanent strand. Pre-4846 the identical timeline self-healed via `pilot-retry-ready`.
- **Interaction with the PR#4849 review's D1 ordering race**: restart in the close→persist window → the SDK poller preflight-declines the fresh fix issue BEFORE the external-close tick lands `pilot-failed`; `ReactToDeclinedFixIssue` → `reactToDeadFixIssue` skips at owner_death.go:144-149 ("source issue not currently designated"), consuming the one-shot reaction; the fallback then designates the declined owner. 4846 converts this race's outcome from benign (retry-ready arrives anyway) to permanent strand.
- **Review-path residue**: `CreateReviewIssue` records the durable claim after issue creation (feedback_loop.go, GH-4841 hunk) — a crash between `CreatePilotIssue` and `ClaimSpawnedFix` leaves an orphan review issue and a duplicate revision issue on restart (no dedup gate).

Gate: requires PR#4846 merged (it is — `classifyOwnerHealth` and the claim fallback are both on main).

## Implementation

1. **Health-check the fallback.** In the `notifyExternalClose` durable-claim fallback (controller.go, GH-4841 hunk), after `fixIssue > 0`: `GetIssue` + `classifyOwnerHealth` (both on main post-#4849). ownerDead → fall through to retry-ready + `emitOwnerDeathAlert`; API error → trust the claim (fail-open, mirroring the GH-4842 dedup path).
[...]

PR #4853, opened 2026-08-11
fix(autopilot): confirm-poll CI status before declaring wait timeout on adopted PRs

### LINKED ISSUE #4851: fix(autopilot): CI wait fires timeout blind on adopted PRs — confirm-poll before deadline, evidence-seeded clock, TerminalLabel on timeout
## Context

2026-08-11 incident: the reconciler adopted open PR#4846 (sha fe0f1d5) at 14:35Z; all 7 checks completed green by 14:43Z; the CI wait clock started at 14:57Z; at 15:27Z the daemon declared "CI timeout after 30m0s" and set stage=failed without ever recording a CI read (persisted ci_status=pending is the fingerprint — controller.go:2239 overwrites it on any successful poll). New variant of the GH-4384 family (green checks + 30m timeout), one layer above the #4408 fix: that fix corrected what a poll computes; here no poll result was ever consulted.

Mechanism:

- `handlePRCreated` (internal/autopilot/controller.go:2104-2132) never looks at CI: it unconditionally sets `Stage=StageWaitingCI` and `CIWaitStartedAt=time.Now()` (:2129-2130). For adopted PRs (reconciler `reconcileOrphanPRs` :6126-6180 and startup `ScanExistingPRs` :6042-6105 → `OnPRCreated` :1859-1945, always `StagePRCreated`+`CIPending` :1876-1877) the 30-minute clock starts at adoption time, not CI time.
- `handleWaitingCI` (controller.go:2137-2280) evaluates the deadline (:2152-2157) BEFORE the only CI read (`c.ciMonitor.CheckCI` :2206). The timeout branch performs no confirming status check, writes stage=failed (:2154) + Error (:2155), never updates ci_status, and sets no TerminalLabel.
- Any poll-suppressing window leaves the wall clock running with zero reads, so the next evaluation past the deadline fails blind. Candidate suppressors in the incident window: rate-limit cooldown (:6794-6807, skips all of processAllPRs, capped 20m — matches the observed 22-min gap) · per-PR circuit breaker (:2019-2023) · `GetPullRequest` error `continue` (:6850-6873) · persist-failure loop (GH-4053 precedent documented at :1243) · restart re-zeroing the clock (:2139-2141).
- Close asymmetry: because no TerminalLabel is set, a later human CLOSE of the stranded PR routes `notifyExternalClose` (:7318) to reclassify the completed execution as failed and default the issue to `pilot-retry-ready` — re-dispatching shipped work. (Manual MERGE is safe: :7022-7108 clears retry labels.)

## Implementation

1. **Confirm-poll before declaring timeout.** In `handleWaitingCI`, do not honor the deadline until at least one `CheckCI` read for the current head SHA has been evaluated in the same tick — either move the deadline check after the status read, or perform one confirming `CheckCI` inside the timeout branch and only fail if it does not return success. A deadline-exceeded tick whose read returns CISuccess must proceed to StageCIPassed.
2. **Seed the wait clock from evidence on the adoption path.** When `OnPRCreated` adopts an existing PR (both reconciler and startup-scan callers), either seed `CIWaitStartedAt` from observable evidence (e.g. PR updated_at), or have `handlePRCreated` resolve CI once inline before entering StageWaitingCI. State the chosen approach in the PR description.
3. **Stamp a TerminalLabel on the timeout branch** so an external close after a bogus timeout cannot default to `pilot-retry-ready` (see existing TerminalLabel sites at controller.go:2619/:3002 for idiom).
4. **Persist what was read**: the timeout branch must record the last-known ci_status (or an explicit never-read marker) instead of leaving the adoption-time `CIPending` default — the persisted row must distinguish "polled and pending" from "never successfully polled".

Out of scope: the unguarded `execution_events` append (controller.go:1701) that recorded a bogus terminal failed rung on a completed execution — separate residue, do not widen scope.

## Acceptance

- Regression test: a PR adopted with all checks already completed green, first `handleWaitingCI` evaluation occurring after the deadline has elapsed → reaches StageCIPassed, not StageFailed (drive with deadline exceeded + CheckCI returning success; must fail on current main).
- Timeout can only be declared after a same-tick CI read that did not return success.
- Timeout branch sets a TerminalLabel; test that an external close after timeout does not arm `pilot-retry-ready`.
- GH-4384 / GH-4415 / GH-4478 / GH-4646 family tests unchanged and passing.

## Refs

- Incident: PR#4846 strand (TASK-467) — adopted 14:35Z, checks green 14:43Z, wait started 14:57Z, bogus timeout 15:27Z, ci_status=pending fingerprint.
- Prior family: GH-4384 (#4408 fix, ci_monitor.go:451-487) · GH-4478 · GH-4646 · GH-4415 restart variant · GH-4053 reconciler-loop precedent (controller.go:1243 comment).

alekspetrov:
## Summary

Fixes GH-4851: autopilot's CI-wait logic was declaring "CI timeout" on adopted PRs without ever polling CI.

**Incident (PR#4846, 2026-08-11):** the reconciler adopted an open PR whose checks had already gone green; the wait clock started blind via `time.Now()` at adoption time; a suppressed processing window delayed the first `handleWaitingCI` evaluation until after the 30-minute deadline had already elapsed. The deadline check ran unconditionally *before* the function's only CI read, so it declared timeout without ever consulting CI — leaving the persisted `ci_status` stuck at its adoption-time `CIPending` default and no `TerminalLabel` set. A later external close of the stranded PR then defaulted the issue to `pilot-retry-ready`, re-dispatching already-shipped work.

## Changes

1. **Confirm-poll before declaring timeout.** `handleWaitingCI` no longer honors an expired deadline until at least one same-tick `CheckCI` read has been evaluated. A read that resolves `CISuccess`/`CIFailure`/`CIConfigMismatch` always wins over an expired clock — extracted into `applyCIOutcome`, which the deadline branch now falls through to only when the read is still `CIPending`/`CIRunning` ("genuinely unresolved").
2. **Seed the wait clock from evidence on adoption.** Chosen approach: seed `CIWaitStartedAt` from the PR's own `updated_at` (via new `seedAdoptedCIWaitClock`, called from both `reconcileOrphanPRs` and `ScanExistingPRs`) rather than resolving CI inline in `handlePRCreated`. Rejected the inline-resolve option because it would add an unconditional extra `CheckCI` call to every PR-creation path (not just adopted ones), risking regressions across the many existing tests that don't mock CI endpoints for arbitrary SHAs. The evidence-seed approach is scoped only to the two adoption call sites and needs no change to `OnPRCreated`'s public signature (which is bound as a callback type by the Linear and Azure DevOps pollers).
3. **TerminalLabel on confirmed timeout**, mirroring the existing idiom in `handleCIFailed` — prevents a later external close from defaulting the issue to `pilot-retry-ready`.
4. **Persist what was actually read.** Because the timeout branch is now only reachable after a confirmed same-tick poll, `CIStatus`/`LastChecked` are always freshly set before a timeout is declared — distinguishing "polled and pending" from "never successfully polled" for free.

Out of scope (per issue): the unguarded `execution_events` append at controller.go:1701 — separate residue, not widened here.

## Test plan

- [x] New regression tests in `internal/autopilot/gh4851_ci_wait_confirm_poll_test.go`:
  - deadline-exceeded + same-tick success → `StageCIPassed` (fails on main)
  - deadline-exceeded + same-tick failure → `StageCIFailed`
  - deadline-exceeded + same-tick still-pending → confirmed `StageFailed`, `TerminalLabel` set, real `ci_status`/`LastChecked` persisted
  - external close after a confirmed timeout does not arm `pilot-retry-ready`
  - reconciler / startup scan seed `CIWaitStartedAt` from PR `updated_at`, not `time.Now()`
  - full end-to-end repro of the PR#4846 shape: adopt stale-but-green PR past deadline → `StageCIPassed`
- [x] `go build ./...`
- [x] `go test ./internal/autopilot/...` (full package, includes GH-4384/GH-4415/GH-4478/GH-4646/GH-4826 families)
- [x] `go vet ./...`

### DIFF
--- a/.agent/tasks/gh-4851.md
+++ b/.agent/tasks/gh-4851.md
@@ -1,0 +1,42 @@
+# GH-4851
+
+**Created:** 2026-08-11
+
+## Problem
+
+GitHub Issue GH-4851: fix(autopilot): CI wait fires timeout blind on adopted PRs — confirm-poll before deadline, evidence-seeded clock, TerminalLabel on timeout
+
+## Context
+
+2026-08-11 incident: the reconciler adopted open PR#4846 (sha fe0f1d5) at 14:35Z; all 7 checks completed green by 14:43Z; the CI wait clock started at 14:57Z; at 15:27Z the daemon declared "CI timeout after 30m0s" and set stage=failed without ever recording a CI read (persisted ci_status=pending is the fingerprint — controller.go:2239 overwrites it on any successful poll). New variant of the GH-4384 family (green checks + 30m timeout), one layer above the #4408 fix: that fix corrected what a poll computes; here no poll result was ever consulted.
+
+Mechanism:
+
+- `handlePRCreated` (internal/autopilot/controller.go:2104-2132) never looks at CI: it unconditionally sets `Stage=StageWaitingCI` and `CIWaitStartedAt=time.Now()` (:2129-2130). For adopted PRs (reconciler `reconcileOrphanPRs` :6126-6180 and startup `ScanExistingPRs` :6042-6105 → `OnPRCreated` :1859-1945, always `StagePRCreated`+`CIPending` :1876-1877) the 30-minute clock starts at adoption time, not CI time.
+- `handleWaitingCI` (controller.go:2137-2280) evaluates the deadline (:2152-2157) BEFORE the only CI read (`c.ciMonitor.CheckCI` :2206). The timeout branch performs no confirming status check, writes stage=failed (:2154) + Error (:2155), never updates ci_status, and sets no TerminalLabel.
+- Any poll-suppressing window leaves the wall clock running with zero reads, so the next evaluation past the deadline fails blind. Candidate suppressors in the incident window: rate-limit cooldown (:6794-6807, skips all of processAllPRs, capped 20m — matches the observed 22-min gap) · per-PR circuit breaker (:2019-2023) · `GetPullRequest` error `continue` (:6850-6873) · persist-failure loop (GH-4053 precedent documented at :1243) · restart re-zeroing the clock (:2139-2141).
+- Close asymmetry: because no TerminalLabel is set, a later human CLOSE of the stranded PR routes `notifyExternalClose` (:7318) to reclassify the completed execution as failed and default the issue to `pilot-retry-ready` — re-dispatching shipped work. (Manual MERGE is safe: :7022-7108 clears retry labels.)
+
+## Implementation
+
+1. **Confirm-poll before declaring timeout.** In `handleWaitingCI`, do not honor the deadline until at least one `CheckCI` read for the current head SHA has been evaluated in the same tick — either move the deadline check after the status read, or perform one confirming `CheckCI` inside the timeout branch and only fail if it does not return success. A deadline-exceeded tick whose read returns CISuccess must proceed to StageCIPassed.
+2. **Seed the wait clock from evidence on the adoption path.** When `OnPRCreated` adopts an existing PR (both reconciler and startup-scan callers), either seed `CIWaitStartedAt` from observable evidence (e.g. PR updated_at), or have `handlePRCreated` resolve CI once inline before entering StageWaitingCI. State the chosen approach in the PR description.
+3. **Stamp a TerminalLabel on the timeout branch** so an external close after a bogus timeout cannot default to `pilot-retry-ready` (see existing TerminalLabel sites at controller.go:2619/:3002 for idiom).
+4. **Persist what was read**: the timeout branch must record the last-known ci_status (or an explicit never-read marker) instead of leaving the adoption-time `CIPending` default — the persisted row must distinguish "polled and pending" from "never successfully polled".
+
+Out of scope: the unguarded `execution_events` append (controller.go:1701) that recorded a bogus terminal failed rung on a completed execution — separate residue, do not widen scope.
+
+## Acceptance
+
+- Regression test: a PR adopted with all checks already completed green, first `handleWaitingCI` evaluation occurring after the deadline has elapsed → reaches StageCIPassed, not StageFailed (drive with deadline exceeded + CheckCI returning success; must fail on current main).
+- Timeout can only be declared after a same-tick CI read that did not return success.
+- Timeout branch sets a TerminalLabel; test that an external close after timeout does not arm `pilot-retry-ready`.
+- GH-4384 / GH-4415 / GH-4478 / GH-4646 family tests unchanged and passing.
+
+## Refs
+
+- Incident: PR#4846 strand (TASK-467) — adopted 14:35Z, checks green 14:43Z, wait started 14:57Z, bogus timeout 15:27Z, ci_status=pending fingerprint.
+- Prior family: GH-4384 (#4408 fix, ci_monitor.go:451-487) · GH-4478 · GH-4646 · GH-4415 restart variant · GH-4053 reconciler-loop precedent (controller.go:1243 comment).
+
+## Acceptance Criteria
+
--- a/internal/autopilot/controller.go
+++ b/internal/autopilot/controller.go
@@ -1944,6 +1944,41 @@ func (c *Controller) OnPRCreated(prNumber int, prURL string, issueNumber int, he
 	}
 }
 
+// seedAdoptedCIWaitClock seeds CIWaitStartedAt for a just-adopted PR (see
+// reconcileOrphanPRs / ScanExistingPRs) from GitHub's own last-activity
+// timestamp instead of letting handlePRCreated default it to time.Now() once
+// the pipeline reaches StageWaitingCI.
+//
+// GH-4851: an adopted PR can have every check long completed by the time
+// OnPRCreated fires — starting the clock at "now" instead of the PR's own
+// last-known-activity time means a suppressed processing window (rate-limit
+// cooldown, circuit breaker, restart) that delays the first handleWaitingCI
+// tick makes the CIWaitTimeout deadline look freshly exceeded, when CI in
+// reality finished the moment the PR stopped changing (PR#4846: adopted
+// 14:35Z, checks green by 14:43Z). This is a lower-bound approximation, not
+// a CI timestamp itself — the deadline-confirmation logic added to
+// handleWaitingCI in the same fix is what actually prevents a false timeout;
+// this just makes the clock (and CIWaitDuration metrics) reflect reality
+// more closely. Only applied while the PR is still sitting in StagePRCreated
+// with a zero CIWaitStartedAt, so it never clobbers a wait already in
+// progress (e.g. a concurrent registration that raced ahead of this call).
+func (c *Controller) seedAdoptedCIWaitClock(prNumber int, updatedAt time.Time) {
+	if updatedAt.IsZero() {
+		return
+	}
+	c.mu.RLock()
+	prState, ok := c.activePRs[prNumber]
+	c.mu.RUnlock()
+	if !ok {
+		return
+	}
+	prState.mu.Lock()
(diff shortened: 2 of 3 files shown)


alekspetrov:
## Post-merge review — PR #4853 (GH-4851, confirm-poll CI status before wait timeout)

**Verdict: HOLDS WITH DEFECTS**

The core fix is correct, matches all four spec items, and the regression suite is genuinely behavioral — all 7 new tests verified failing against pre-fix main (`319dc230` controller.go swapped in under the merged test file: 7/7 FAIL; restored: full `./internal/autopilot` package green, `-race` clean). The defects below are second-order interactions, not breaks in the primary incident fix.

### Spec compliance

1. **Confirm-poll before timeout** — ✅. Deadline is captured at controller.go:2208 but only honored at :2307 after the same-tick `CheckCI` at :2257 returned a non-error read that `applyCIOutcome` (:2329) classified as still-unresolved. Success/failure/config-mismatch all win over an expired clock. The read is against the refreshed head SHA (:2230-2244), per spec.
2. **Evidence-seeded clock** — ✅. `seedAdoptedCIWaitClock` (:1965) from PR `updated_at`, called at both adoption sites (:6205, :6287); `handlePRCreated` preserves a non-zero seed (:2170-2172). PR description states the chosen approach and the rejected alternative, as required.
3. **TerminalLabel on timeout** — ✅ stamped (:2318), mirrors the handleCIFailed idiom. But see D1/D2.
4. **Persist what was read** — ✅ structurally free: the timeout branch is only reachable after `prState.CIStatus`/`LastChecked` were freshly written (:2290-2291), and the `StageFailed` early-return in ProcessPR (:2094-2096) only applies on *subsequent* ticks, so the timeout tick itself persists via :2131. `LastChecked` non-zero is the "actually polled" fingerprint; test asserts it.

Acceptance: green-adopted-past-deadline → StageCIPassed ✅ (fails pre-fix ✅); timeout only after non-success same-tick read ✅; external-close-after-timeout test exists and asserts on real POSTed labels ✅; GH-4384/4415/4478/4646 family files untouched (PR touches exactly 3 files) and passing ✅.

### Defects

**D1 — MAJOR — confirmed-timeout + un-reset wait clock creates a new terminal-stranding class on the CI re-trigger paths.**
`internal/autopilot/controller.go:2902` (infra-outage rerun), `:5318` (auto-rebase), `:5434` (mechanical go.mod resolution) all re-enter `StageWaitingCI` **without resetting `CIWaitStartedAt`** (contrast :5709-5710 and :5781-5782, which do). Concrete scenario: PR waits 25m, CI fails, classified as infra outage, jobs rerun → Stage back to WaitingCI with the original clock; the rerun is `in_progress` on the next tick at minute 31 → `deadlineExceeded` true, same-tick read = pending → **instant "confirmed" timeout**, now stamped `TerminalLabel=pilot-failed`. Pre-PR this same tick blind-timed-out too, but the external close then re-armed `pilot-retry-ready` (noisy, self-healing). Post-PR the close routes to `pilot-failed` with no spawned fix issue and no owner — permanent strand unless a human intervenes. The confirm-poll cannot distinguish "genuinely waited 30m unresolved" from "clock never reset by an intermediate re-entry"; stamping a terminal label on that ambiguity converts a pre-existing nuisance bug into a stranding bug. Fix: reset `CIWaitStartedAt = time.Now()` at the three re-entry sites (they explicitly say "triggers new CI").

**D2 — MEDIUM — TerminalLabel is not persisted, and unlike every prior setter, the timeout branch does not close the PR itself.**
`SavePRState`/`LoadAllPRStates` (internal/autopilot/state_store.go:638-700, :731-792) omit `terminal_label`; types.go:1166-1168 documents the field as in-memory-only, "safe because a restart re-enters the handler that sets it before the PR can reach notifyExternalClose again." That invariant is **false for this branch**: the timed-out PR stays open on GitHub indefinitely at StageFailed, and the human close may come days and several daemon restarts later. Structural mitigation that mostly saves it: `RestoreState` skips StageFailed rows (controller.go:1769), so post-restart the still-open PR is untracked and `reconcileOrphanPRs` re-adopts it within 60s — a green PR then heals to CIPassed; a still-hung one re-confirms the timeout and re-stamps the label. Residual hole: a close landing in the re-adoption window (StagePRCreated/WaitingCI, before the re-confirming tick) reaches `notifyExternalClose` with an empty TerminalLabel and no spawned-fix claim → `pilot-retry-ready` re-armed — the incident shape, minutes-wide, restart-gated. Either persist the field (column exists-adjacent precedent: `escalation_reason`) or record a durable claim analog for timeouts; at minimum update the stale types.go comment.

**D3 — MINOR (pre-existing, adjacent) — the consecutive-API-failure branch is the remaining door to the exact incident fingerprint.**
controller.go:2267-2275: five consecutive `CheckCI` errors → StageFailed with **zero successful polls**, `ci_status` left at the adoption-time default, `LastChecked` zero, **no TerminalLabel** → a later external close defaults to retry-ready. This is "stage=failed, ci_status=pending, never successfully polled" — the very row-shape #4851 calls the fingerprint. Not the timeout branch, so technically out of spec scope, and untouched by this PR — but worth a follow-up since it defeats spec item 4's distinguishability promise in the error case.

### Interaction checks (requested)

- **Compose with #4846 durable-claim fallback / #4854 owner-death hardening:** `notifyExternalClose` ladder on merged main is (1) pilot-done skip :7526, (2) human-recovery-PR skip :7544, (3) `TerminalLabel` if set :7571, (4) GH-4841 durable spawned-fix claim + GH-4852 owner-health check :7574-7619, (5) default retry-ready. 4853's label wins over the claim; a single `issueLabel` variable makes double-designation (failed + retry-ready) impossible. One narrow semantic gap: a set TerminalLabel bypasses #4854's owner-health check, so a PR that earlier spawned a fix issue (claim recorded), was re-driven to WaitingCI, and then confirm-timed-out will, on close, label its source pilot-failed *without* the dead-owner re-arm even if that fix issue died. Note-level; requires the D1 path to be live first.
- **Merge-ordering with #4854:** no textual overlap (4853: handleWaitingCI/handlePRCreated/adoption sites; 4854: notifyExternalClose claim-block interior, owner_death.go, feedback_loop.go, state_store.go). Semantics compose as designed; full package green on the merged union.
- **Liveness / no unbounded wait:** CheckCI erroring every tick → StageFailed after 5 strikes (:2267), bounded (see D3 for its label gap). Rate-limit cooldown skips whole ticks (capped 20m, :6905-6913); after it lapses the next tick confirm-polls and can declare the timeout — no spin. Only unbounded residue: `GetPullRequest` erroring forever with an empty stored SHA (:2223-2224) retries silently without striking the failure counter — pre-existing, rare (adopted PRs always carry a SHA).

### Notes / nits

- Tests are behavioral: driven through `ProcessPR`/`reconcileOrphanPRs`/`ScanExistingPRs` against httptest GitHub servers; the close test asserts on labels actually POSTed. The ConfirmedPending test's `CIStatus == CIPending` assertion is vacuous on its own (equals the adoption default) — the test comment admits this and correctly leans on `LastChecked`; fine.
- `seedAdoptedCIWaitClock`'s seed is in-memory until the first ProcessPR persist; a restart inside that <60s window drops it and the clock falls back to `time.Now()`. Harmless — the confirm-poll is the real guard, as the function's own comment says.
- `.agent/tasks/gh-4851.md` ships with an empty trailing `## Acceptance Criteria` header (template residue); project convention also says task docs move to `archive/` once merged.
- The comment at :2317 cites "controller.go:2619" — already stale after #4854's shifts; prefer symbol references.

**Bottom line:** primary incident fix (PR#4846 shape) is solid, tested, and mutation-verified. File a follow-up for D1 (clock reset at the three WaitingCI re-entry sites) before the next infra-outage/auto-rebase event turns a recoverable timeout into a stranded pilot-failed; D2's persistence gap and D3 can ride along in it.

PR #4865, opened 2026-08-13
fix(observability): expose the running daemon version on health/metrics/doctor

### LINKED ISSUE #4864: fix(observability): expose the running daemon version — hot restarts are invisible to every status surface
## Context

The self-upgrade restart leg works — `syscall.Exec` in place (`internal/upgrade/restart.go:78`), verified at boot by the GH-3600 reconcile (`internal/upgrade/reconcile.go:48`, logged `upgrade verified complete … via "hot restart"` on every release train 07-31 → 08-12). But **no status surface reports what the running process is**, so three operator sessions (08-08 → 08-13) misdiagnosed healthy hot-restarted daemons as "installed but never restarted", culminating in an unnecessary operator restart on 08-13. The signal set is structurally blind:

- Board `ver` shells the **disk** binary (`/usr/local/bin/pilot version` → compiled-in `version` var of the just-installed file, `cmd/pilot/commands.go:438`).
- Board `uptime` is `ps` etime — **survives `syscall.Exec`** (PID unchanged), so it cannot detect a hot restart.
- Gateway `/health` hardcodes `"version": "0.1.0"` (`internal/gateway/server.go:557`).
- Metrics (:9091) have no version gauge.
- `pilot doctor` staleness check compares the **invoked** binary's version (`internal/health/health.go:416-454`) — run on the box it reads the new disk binary and reports "up to date" regardless of the process.
- The staleness alert counts `ReleasesBehind` ≥ 3 (`cmd/pilot/main.go:3735`) — a genuinely stale process 1 patch behind would never alert either.

## Implementation

1. **`pilot_build_info` metric**: register a gauge `pilot_build_info{version="<version>", commit="<vcs rev>"} 1` on the :9091 registry at daemon startup, sourced from the running process's compiled-in `version` var (same source as the TUI banner, `internal/dashboard/tui.go:1555`). Standard build-info idiom; the value changes across a hot restart with no PID change.
2. **Real `/health` version**: replace the hardcoded `"0.1.0"` in `internal/gateway/server.go:557` with the real version (plumb the `version` var; do not shell out).
3. **`pilot doctor` disk-vs-running check**: extend the staleness check to fetch the running daemon's version from the local metrics endpoint (config port, default :9091) or `/health`, and compare THREE values: running process vs disk binary vs latest release. Report `running ≠ disk` explicitly ("hot restart pending or failed — check daemon.log for 'upgrade verified complete'"). Degrade gracefully (skip with a note) when no daemon is reachable.

## Acceptance

- `curl -s localhost:9091/metrics | grep pilot_build_info` returns the running version + commit; unit test asserts the gauge is registered and labeled from the version var.
- `/health` returns the real version; test updated (no `"0.1.0"` literal left in `internal/gateway/`).
- Doctor: table-driven tests for the three-way comparison — running=disk (ok), running≠disk (explicit mismatch wording), daemon unreachable (graceful skip). No change to upgrade-flow behavior.
- Existing upgrade/reconcile suites pass unchanged.

## Out of scope (recorded, not this issue)

Latent defects found during the same research, none implicated in a live failure: `state.Save` is the last fallible step AFTER the binary is committed (`internal/upgrade/graceful.go:137-139` — an error there aborts before the restart leg); repeat-attempt backup poisoning (`internal/upgrade/upgrade.go:458-482` copies the already-new disk binary over `.backup`); `ReleasesBehind` threshold is count-based, blind to age. Operator follow-up once shipped: point `pilot-board-remote`'s `ver` at the build-info metric instead of the disk binary.

## Refs

- Navigator: `.agent/tasks/TASK-476-running-version-observability.md` · memory `hot-restart-preserves-pid-uptime-false-mismatch` (pitfall).
- Lineage: GH-3600 boot self-verify · #4470 writable-dir preflight · #4683 drain-pause — the upgrade flow itself is healthy; this issue is purely the observation surface.

alekspetrov:
## Summary

Closes GH-4864. The self-upgrade hot restart (`syscall.Exec`) works correctly, but every status surface was blind to it — disk-binary `version`, PID-based `uptime`, a hardcoded `/health` version, no metrics gauge, and a doctor check that only compared disk-vs-latest. Three operator sessions (08-08 → 08-13) misdiagnosed healthy hot-restarted daemons as stuck.

- `pilot_build_info{version,commit}` gauge registered on the `:9091` Prometheus registry, sourced from the running process's compiled-in `version` var (commit parsed from the git-describe version string, no shelling out).
- `/health` and `/api/v1/status` now report the real running version instead of the hardcoded `"0.1.0"`.
- `pilot doctor` gains a `self-upgrade.running-vs-disk` check that fetches the live daemon's `/health` and explicitly flags a running≠disk mismatch ("hot restart pending or failed — check daemon.log for 'upgrade verified complete'"), degrading gracefully when no daemon is reachable locally.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/gateway/... ./internal/health/...` — all pass
- [x] Table-driven tests cover `checkRunningVsDiskVersion`: match, mismatch (with Fix hint), unreachable, port unconfigured, non-200, unparseable body, `v`-prefix normalization
- [x] `pilot_build_info` gauge tested for registration + version/commit labeling, order-independence of `SetVersion`/`SetMetricsSource`
- [x] No `"0.1.0"` literal remains in `internal/gateway/`


alekspetrov:
## Post-merge review — PR#4865 (GH-4864, squash 7660794f)

**Verdict: APPROVE** — all three spec legs land where the issue aimed them: the gauge is on the exact exporter that serves the daemon's `/metrics`, `/health` and `/api/v1/status` are plumbed from the compiled-in `main.version` var (no shell-out anywhere), and doctor's `running-vs-disk` check reads the live daemon over the config-driven gateway port. Three mutants killed. Two minor defects (one operator-facing wording gap that briefly masks the check's own purpose, one docs regression), three notes.

### Spec-leg verification (isolated worktree @ merged main)

1. **Gauge on the right registry, right source.** `Server.SetVersion` (server.go:347) → `PrometheusExporter.SetBuildInfo` on `s.prometheusExporter` — the same exporter `handleMetrics` (server.go:938) reads, registered on the same mux as `/health` (server.go:237/:240), i.e. the one HTTP server the daemon binds. Not a fresh default registry — this codebase's hand-rolled text exporter has no global registry to accidentally register against, and `writeGaugeLabeled` emits inside `WritePrometheus` (prometheus.go:296-306). Source is the ldflags-stamped `version` var (cmd/pilot/main.go:60, Makefile `-X main.version`); `commitFromVersion` recovers the commit from the git-describe suffix instead of adding a second ldflag — reasonable, and the "unknown" fallback for clean-tag/no-ldflags builds is tested. Order-independence of `SetVersion`/`SetMetricsSource` is real (SetMetricsSource re-applies build info, server.go:335) and pinned by its own two-way test.
2. **`/health` plumbed, no literal left.** `grep '0.1.0' internal/gateway/` (non-test): zero hits. Both daemon entrypoints wire it: webhook mode `p.Gateway().SetVersion(version)` (main.go:1254) and polling mode `gwServer.SetVersion(version)` (main.go:2590) — the only two `NewServerWithAuth` construction sites outside the package (main.go:2588, pilot.go:493, the latter reached via the former). `/health` sits on the public no-auth mux, so the GH-4784 bearer-auth path can't 401 doctor's probe; the unset case (`""`) is pinned by `TestHealthEndpoint_VersionUnset`.
3. **Doctor three-way, config port, graceful degrade.** `checkRunningVsDiskVersion` (health.go:459) does a real GET of `http://127.0.0.1:{cfg.Gateway.Port}/health` — port from config (health.go:559-563), no hardcoded 9091; :9091 on the box is just the box's configured gateway port, and since `/metrics` and `/health` share that server, implementing the spec's "metrics endpoint or /health" as /health-only is equivalent and simpler. Skips (StatusOK + note) on port≤0 (no network call — pinned by a `t.Fatal` getter), connection error, non-200, and unparseable body; mismatch is StatusWarning with the "hot restart pending or failed" wording plus a Fix pointing at the `upgrade verified complete` reconcile marker — actionable, matches the incident shape. Disk-vs-latest (`checkSelfUpgradeStaleness`) runs alongside, so the three values are all visible. `pilot doctor` passes its own compiled-in version as "disk" (doctor.go:33) — correct, doctor *is* the disk binary when run from the install path, and the function comment says so.
4. **Upgrade flow untouched.** `git diff 7660794f^..7660794f -- internal/upgrade/` is empty; `go test ./internal/upgrade/...` green (9.9s).

### Mutation checks (code restored after each)

- **M1** — reverted `/health` to hardcoded `"0.1.0"` → KILLED: `TestHealthEndpoint` ("Expected version 'v2.259.1-29-g35450fea', got '0.1.0'") *and* `TestHealthEndpoint_VersionUnset` both fail — the acceptance criterion's literal is load-bearing from two directions.
- **M2** — neutered the mismatch branch (`if running == disk || true`) → KILLED: table case "running differs from disk" (status + message) and `TestCheckRunningVsDiskVersion_MismatchHasFix` (Fix hint) both fail.
- **M3** — dropped the `SetBuildInfo` re-application inside `SetMetricsSource` (server.go:335) → KILLED: the "SetVersion before SetMetricsSource" subtest fails on the missing labeled gauge — the order-independence seam is genuinely pinned, not incidental.
- Full `go test ./internal/gateway/ ./internal/health/` green (2.0s / 7.9s).

### Defects

**D1 — minor — internal/health/health.go:502-505 — a reachable daemon that reports no version is called "not reachable", which masks the check's own target during the first upgrade window.** Every degrade leg reuses one message: `"daemon not reachable locally, skipping running-version check"`. But the `payload.Version == ""` leg fires precisely when the daemon *is* reachable and answering 200 — which is what every pre-4865 daemon does (`/health` had no version field until this PR). Concretely: during the first post-merge release train, the running process is a pre-4865 build, disk is new — the exact running≠disk state this check was built to expose — and doctor reports an OK "not reachable, skipping". Self-resolving after one successful hot restart, but the wording stays wrong for the non-200 and unparseable legs forever (a 503ing live daemon is not "not reachable"). Follow-up shape: distinct message for reachable-but-versionless ("daemon reachable but reports no version (pre-GH-4864 build) — cannot compare; treat disk-vs-latest as the only signal"), which would also have made the transitional window visible instead of silent.

**D2 — trivial (docs) — .agent/system/FEATURE-MATRIX.md:3 — "Last Updated" regressed 2026-08-12 → 2026-08-11** while adding the 08-13 row (stale-base artifact; #4863 had already bumped it). Merged main now claims the matrix is two days older than its newest entry. One-line fix for the next docs pass.

### Notes

- N1: `pilot_build_info` existence is coupled to autopilot metrics wiring — the exporter is only constructed by `SetMetricsSource`, which both entrypoints gate on autopilot controllers existing (main.go:1261-1262, :2634-2636). A daemon with autopilot disabled 503s on `/metrics` and exposes no build info there. Pre-existing `/metrics` shape (nothing else is served without a source either), `/health` still carries the version, and the founder box runs autopilot — ledger only, but worth remembering if the spec's "at daemon startup" is ever read literally for a metrics-only consumer (e.g. the planned `pilot-board-remote` `ver` pointer).
- N2: the doctor probe hardcodes `127.0.0.1` and ignores `cfg.Gateway.Host` — a gateway bound to a specific non-loopback interface would silently skip. Default host is loopback and the box binds loopback + tunnel; cosmetic.
- N3: skips render as StatusOK, so "daemon not reachable, skipping" shows as a green check line. Consistent with the acceptance's "graceful skip", and a warning would false-alarm every laptop `pilot doctor` run — right call, noting the tradeoff for the ledger. Also: the check only runs when `currentVersion != ""` (health.go:549), which is always true in practice given the `"1.0.0"` fallback.

**Bottom line**: the observation surface the three misdiagnoses needed now exists and is tested at the seams that matter; nothing touches upgrade behavior. D1 is the one real follow-up candidate — cheap, and it converts the feature's own first deployment from a silent skip into a visible signal.

Issue #4716, opened 2026-08-04
feat(executor): post-task invariant tripwire sweep on ExecutionLifecycle.Finish — root-clean, labels, children, worktree; alert-never-block (TASK-441 L5)


alekspetrov:
## Context

TASK-441 Leg 5 (`.agent/tasks/TASK-441-contract-hardening-tune-up.md`) — the catch-all. A week of incidents shared one shape: wiring bugs wearing a green test suite (self-review executing in the repo root for months, labels never applied for 19 days, epic-decompose discarding child work). Point fixes closed each instance; this leg makes any *new* call site reintroducing a past class get caught on first fire, not months later.

TASK-404 made `ExecutionLifecycle.Finish` (`internal/executor/lifecycle.go:283`) the universal terminal transition — every execution path, including future ones, funnels through it. That is the one place a post-task invariant sweep covers everything.

Leg 2's dead-man primitive is now merged (PR#4712): `alerts.DeadManTracker` / `Engine.RegisterDeadManTracker` (`internal/alerts/deadman.go`). Each tripwire check feeds one.

## Acceptance

- Optional post-`Persist` hook in `ExecutionLifecycle.Finish`, terminal paths only, running an invariant sweep:
  - **(a) root-clean** — no staged/unstaged diff in `task.ProjectPath` after the task (pitfall `runselfreview-runs-in-repo-root-phantom-reimplementation`: misrouted subprocess staged a phantom reimplementation in the shared root).
  - **(b) label lifecycle completed** — the task's in-progress marker was applied and terminally resolved (pitfall `poller-labels-removed-log-means-never-applied`).
  - **(c) decomposed children all terminal** — no orphaned non-terminal children when the parent finishes.
  - **(d) worktree pruned + no commits-without-PR** — pitfall `pitfall_epic_decompose_discards_child_work`: worktree cleanup discarded uncollected child commits.
- **Log-and-alert only — never block or fail the Finish path.** A sweep panic/error must be recovered and logged; the terminal transition always completes.
- Each check registers a `DeadManTracker` (attempt on every sweep, success on pass, failure with metadata on violation) so a *streak* of violations pages via the engine's existing dispatch.
- Sweep is cheap (target: no network calls in the hot path; git status inspection + DB reads only). Checks that need external state (label lifecycle) read what the execution already recorded, not the tracker API.
- Table-driven tests: each check pass/fail, hook never propagates errors, non-terminal transitions skip the sweep.
[...]

PR #4867, opened 2026-08-13
fix(alerts): dead-man streak alerts can never fire (GH-4866)

### LINKED ISSUE #4866: fix(alerts): dead-man streak alerts can never fire — daemon rule set omits all dead-man rules (kill-drill A failure)
## Context

TASK-441 kill-drill A (2026-08-13, box v2.259.1): a sandbox repo was archived so every GitHub mutation 403'd. 57 minutes of continuous seam failures — 12+ `Failed to notify task started (SDK path)` WARNs, `failed to strip pilot-in-progress label` ERRORs, push-retry exhaustion, ~36 infra-failed executions — produced **zero** dead-man output. Generic per-task `task_failed` alerts fired throughout (engine + channels healthy), proving the gap is rule wiring, not the engine.

Root cause chain (static-provable):

1. **The daemon's rule set never contains any dead-man rule.** The four rules (`label_lifecycle_failure_streak`, `self_review_failure_streak`, `finish_tripwire_failure_streak`, `intent_judge_failure_streak`) exist only in `alerts.defaultRules()` (`internal/alerts/types.go:466-521`), whose sole caller is `alerts.DefaultConfig()` → used only by the one-shot `--alerts` CLI (`cmd/pilot/commands.go:666`). The daemon path is `cmd/pilot/main.go:2962` → `getAlertsConfig` (`cmd/pilot/adapters.go:179-232`) → `alerts.FromConfigAlerts` (`internal/alerts/config.go:12-33`), seeded from a DIFFERENT 7-rule list at `internal/config/config.go:543-624` that omits all four. PR#4712/#4724 touched `types.go` only. Live config compounds it: `config.Load` replaces defaults wholesale with any persisted `alerts.rules:` — the deployed config carries a legacy 5-rule list.
2. **Silent no-match.** `handleDeadManStreak` (`internal/alerts/deadman.go:308-321`) loops `e.config.Rules`, matches nothing, returns — no log, no metric. The tracker itself never logs; a threshold crossing with no rule is indistinguishable from nothing happening.
3. **Global streak.** `labelLifecycleDeadManTrackerName = "label_lifecycle"` (`cmd/pilot/handlers.go:942`) is one memoized counter for all 12 projects; repo is metadata only (`handlers.go:842`). Any healthy repo's `RecordSuccess` (`handlers.go:848`) zeroes it (`deadman.go:187`) — with 11 healthy repos polling every 30s, a threshold-10 streak is unreachable. Registration also happens inside the per-issue body (`handlers.go:834-839`), so a nil engine yields a fresh unmemoized tracker per issue.
4. **Unwired seams.** `stripInProgressLabelOnTerminalFailure` (`internal/executor/lifecycle.go:408-459`, ERROR at :450) and push-retry exhaustion (`internal/executor/runner.go:4879/4888`) record nothing into any tracker despite `l.alertProcessor` being available.
5. **Strict-equality fire.** `deadman.go:205`: `if consecutive != t.threshold || t.engine == nil { return }` — fires only at exactly N; one missed increment skips the fire permanently; nil engine drops silently.
6. **Doctor blind.** `internal/health/subsystems.go:59-68` checks only enabled+channels — no rule-coverage check; `checkAlertProcessorWiredOnRunner` (`:83-88`) is stale and hardcoded `return false` (contradicted by `main.go:3050`), a permanently-red check that trains operators to ignore doctor.

## Implementation

1. **Single source of truth for rules**: union rules from `alerts.defaultRules()` into the engine config for any `Type` absent from the user's list (in `alerts.FromConfigAlerts` or `getAlertsConfig`); remove or delegate the duplicate list in `internal/config/config.go:543-624`. This must also repair configs with persisted legacy `alerts.rules:` lists (no migration exists).
2. **Fail loudly on no-match**: `handleDeadManStreak` (and the intent-judge/dispatch-loop handlers) WARN `dead-man streak reached threshold but no alert rule matched` with tracker/alert_type/consecutive_failures when no rule matches; add an unconditional WARN in `RecordFailure` at threshold crossing so streaks are greppable even with alerts disabled. Change the fire condition to `>=` threshold with a fired-once flag reset on success.
3. **Per-repo streak keying**: register `"label_lifecycle:"+owner+"/"+repo` (memoization at `deadman.go:141-149` already handles instances); hoist registration out of the per-issue body.
4. **Wire the unwired seams**: emit `AlertEventTypeDeadManAttempt/Success/Failure` around the label strip in `lifecycle.go:444-455` and on push-retry exhaustion in `runner.go`, via the existing alert processor (nil-safe).
5. **Doctor**: add a rule-coverage check (every handler-emitted AlertType has an enabled rule in the config-derived set); fix or delete the hardcoded-false `checkAlertProcessorWiredOnRunner`.

## Acceptance

- Table test: for every AlertType emitted by engine handlers (enumeration exists in `types_test.go:74-79`), the daemon-path config (`getAlertsConfig`) yields an enabled rule — including when the input config carries the legacy 5-rule list.
- Seam test: 10 consecutive notify failures for repo A with interleaved successes from repo B still fires `label_lifecycle_failure_streak` for repo A (per-repo keying); the alert log line `alert fired rule=label_lifecycle_failure_streak` appears.
- No-match path: with the rule deliberately removed, the new WARN is asserted (kills the silent-return mutation).
- Lifecycle strip failure and push-retry exhaustion each produce dead-man failure events in a test crossing the executor→alerts seam (no argument-discarding mocks, per TASK-441 L1 rules).
- Doctor: legacy 5-rule config → coverage check reports the missing dead-man rules; the stale processor check no longer hardcodes false.
- Existing alerts suites pass; no channel-routing changes.

## Out of scope

Operator re-drill per `.agent/sops/operations/task441-kill-drill.md` (required to tick TASK-441's acceptance box after this merges and reaches the box).

## Refs

- Kill-drill record: `.agent/sops/operations/task441-kill-drill.md` § "Drill record — 2026-08-13"; box daemon.log 11:51–12:49Z.
- TASK-441 (`.agent/tasks/TASK-441-contract-hardening-tune-up.md`) — acceptance box 1 blocked on this.
- Lineage: PR#4712 (DeadManTracker) · PR#4724 (Finish tripwires) · GH-4687/GH-4692 (the "wired to nothing" class this reproduces inside its own countermeasure).

alekspetrov:
## Summary
- Fixes GH-4866: TASK-441 kill-drill A found the daemon's `alerts.rules` config omitted every dead-man rule, so 57 minutes of continuous seam failures (label-strip, push-retry exhaustion, notify-task-started) during a sandboxed-repo drill produced zero dead-man output.
- `alerts.FromConfigAlerts` now unions in any `alerts.defaultRules()` Type missing from the caller's rule list — a persisted or legacy config can no longer silently drop a Type added after it was authored.
- `DeadManTracker.RecordFailure` now logs an unconditional WARN at the threshold crossing (regardless of rule/engine wiring) and fires on a `>=`/fired-once gate instead of exact equality, so trackers that start counting mid-streak still fire; `handleDeadManStreak`/`handleIntentJudgeFailureStreak`/`handleDispatchLoopBreaker` WARN when a streak event has no matching rule at all.
- `label_lifecycle` dead-man tracking is now keyed per-repo (`"label_lifecycle:"+owner/repo`) instead of one global counter, so one broken repo's failures aren't diluted by every other healthy repo's successes.
- Wired the two previously-silent seams the kill-drill exercised: the `pilot-in-progress` label-strip path (`lifecycle.go`) and git-push retry exhaustion (`runner.go`) now emit dead-man attempt/success/failure events through the existing nil-safe `AlertEventProcessor` relay.
- `pilot doctor`: `checkAlertProcessorWiredOnRunner` no longer hardcodes `false` (GH-4716 already fixed the underlying gap); added a new rule-coverage check that flags any handler-emitted `AlertType` with no enabled rule, built on a new shared `config.AlertsConfig.ToAlertConfig` conversion so both the daemon start path and the doctor check go through one implementation.

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` — full suite green
- [x] New table test: `getAlertsConfig` yields full AlertType coverage even from the legacy 5-rule config (`cmd/pilot/wiring_test.go`)
- [x] New doctor coverage-check test (`internal/health/subsystems_test.go`)
- [x] New seam tests for lifecycle label-strip and push-retry-exhaustion dead-man events, using real fake-`gh`/no-remote-repo fixtures (no argument-discarding mocks)
- [x] `make check-secrets` clean

### DIFF
--- a/cmd/pilot/adapters.go
+++ b/cmd/pilot/adapters.go
@@ -175,60 +175,13 @@ func wireProjectAccessChecker(runner *executor.Runner, cfg *config.Config) func(
 	return func() { _ = db.Close() }
 }
 
-// getAlertsConfig extracts alerts configuration from the main config
+// getAlertsConfig extracts alerts configuration from the main config.
+// GH-4866: delegates to config.AlertsConfig.ToAlertConfig so this
+// conversion has exactly one implementation, shared with internal/health's
+// doctor rule-coverage check (which can't reach this function directly —
+// cmd/pilot imports internal/health, not the reverse).
 func getAlertsConfig(cfg *config.Config) *alerts.AlertConfig {
-	if cfg.Alerts == nil {
-		return nil
-	}
-
-	alertsCfg := cfg.Alerts
-
-	// Convert to alerts package types (channel configs are shared types, passed directly)
-	channels := make([]alerts.ChannelConfigInput, 0, len(alertsCfg.Channels))
-	for _, ch := range alertsCfg.Channels {
-		channels = append(channels, alerts.ChannelConfigInput{
-			Name:       ch.Name,
-			Type:       ch.Type,
-			Enabled:    ch.Enabled,
-			Severities: ch.Severities,
-			Slack:      ch.Slack,     // Same type, direct pass-through
-			Telegram:   ch.Telegram,  // Same type, direct pass-through
-			Email:      ch.Email,     // Same type, direct pass-through
-			Webhook:    ch.Webhook,   // Same type, direct pass-through
-			PagerDuty:  ch.PagerDuty, // Same type, direct pass-through
-		})
-	}
-
-	rules := make([]alerts.RuleConfigInput, 0, len(alertsCfg.Rules))
-	for _, r := range alertsCfg.Rules {
-		rules = append(rules, alerts.RuleConfigInput{
-			Name:        r.Name,
-			Type:        r.Type,
-			Enabled:     r.Enabled,
-			Severity:    r.Severity,
-			Channels:    r.Channels,
-			Cooldown:    r.Cooldown,
-			Description: r.Description,
-			Condition: alerts.ConditionConfigInput{
-				ProgressUnchangedFor: r.Condition.ProgressUnchangedFor,
-				ConsecutiveFailures:  r.Condition.ConsecutiveFailures,
-				DailySpendThreshold:  r.Condition.DailySpendThreshold,
-				BudgetLimit:          r.Condition.BudgetLimit,
-				UsageSpikePercent:    r.Condition.UsageSpikePercent,
-				Pattern:              r.Condition.Pattern,
-				FilePattern:          r.Condition.FilePattern,
-				Paths:                r.Condition.Paths,
-			},
-		})
-	}
-
-	defaults := alerts.DefaultsConfigInput{
-		Cooldown:           alertsCfg.Defaults.Cooldown,
-		DefaultSeverity:    alertsCfg.Defaults.DefaultSeverity,
-		SuppressDuplicates: alertsCfg.Defaults.SuppressDuplicates,
-	}
-
-	return alerts.FromConfigAlerts(alertsCfg.Enabled, channels, rules, defaults)
+	return cfg.Alerts.ToAlertConfig()
 }
 
 // qualityCheckerWrapper adapts quality.Executor to executor.QualityChecker interface
--- a/cmd/pilot/handlers.go
+++ b/cmd/pilot/handlers.go
@@ -832,7 +832,7 @@ func handleGithubIssueEventSDK(ctx context.Context, cfg *config.Config, ev sdkco
 		// zero attempts already are (Stale) instead of hiding behind an
 		// error-only counter.
 		labelTracker := alertsEngine.RegisterDeadManTracker(
-			labelLifecycleDeadManTrackerName,
+			labelLifecycleDeadManTrackerName(repoOwner+"/"+repoName),
 			alerts.AlertTypeLabelLifecycleFailureStreak,
 			alerts.DefaultDeadManFailureThreshold,
 			alerts.DefaultDeadManWindow,
@@ -935,11 +935,29 @@ func fetchGithubIssueForSDKTask(ctx context.Context, client *githubSDK.Client, o
 	return issue
 }
 
-// labelLifecycleDeadManTrackerName is the alerts.DeadManTracker registration
-// name (TASK-441 L2, GH-4709) for the notifyTaskStartedSDK call site above —
-// global (not per-repo) since the GH-4687 incident it guards was the whole
-// label-lifecycle path going silent, not a single repo's.
-const labelLifecycleDeadManTrackerName = "label_lifecycle"
+// labelLifecycleDeadManTrackerPrefix is the alerts.DeadManTracker
+// registration name prefix (TASK-441 L2, GH-4709) for the
+// notifyTaskStartedSDK call site above. GH-4866: per-repo, not global — a
+// global shared counter let one repo's real, sustained label-lifecycle
+// failures get diluted by every other (healthy) repo's RecordSuccess calls
+// on the same tracker, so a single broken repo's streak could take far
+// longer to reach threshold (or never reach it at all, on a large
+// multi-repo fleet) than the DefaultDeadManFailureThreshold consecutive
+// failures the tracker is supposed to guarantee. Mirrors
+// sdkPreFlightJudge's per-repo "intent_judge:"+repoFullName keying
+// (poller_github.go) for the same reason.
+const labelLifecycleDeadManTrackerPrefix = "label_lifecycle:"
+
+// labelLifecycleDeadManTrackerName returns the per-repo tracker name for
+// repoFullName ("owner/repo"). Registered once per repo poller at startup
+// (startGithubSDKPollerForRepo, poller_github.go — mirroring the self-review
+// tracker registered there) and resolved again here by the same name on
+// every event, relying on RegisterDeadManTracker's memoize-by-name so both
+// call sites share one set of counters per repo instead of the startup
+// registration and the event-time registration racing to create two.
+func labelLifecycleDeadManTrackerName(repoFullName string) string {
+	return labelLifecycleDeadManTrackerPrefix + repoFullName
+}
 
 // notifyTaskStartedSDK applies the pilot-in-progress label and posts the
 // task-started comment via studio-sdk's tested Notifier (GH-4687). Extracted
--- a/cmd/pilot/label_lifecycle_deadman_test.go
+++ b/cmd/pilot/label_lifecycle_deadman_test.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestLabelLifecycleDeadManTrackerName_PerRepo verifies GH-4866's per-repo
+// keying: two different repos must resolve to two different tracker names,
+// and the same repo must always resolve to the same name (so
+// alerts.Engine.RegisterDeadManTracker's memoize-by-name shares one set of
+// counters for it across the startup registration in poller_github.go and
+// the event-time lookup in handlers.go).
+func TestLabelLifecycleDeadManTrackerName_PerRepo(t *testing.T) {
+	nameA := labelLifecycleDeadManTrackerName("qf-studio/pilot")
+	nameB := labelLifecycleDeadManTrackerName("qf-studio/other-repo")
+
+	if nameA == nameB {
+		t.Fatalf("expected distinct tracker names for distinct repos, got %q for both", nameA)
+	}
+	if got := labelLifecycleDeadManTrackerName("qf-studio/pilot"); got != nameA {
+		t.Errorf("expected stable tracker name for the same repo, got %q want %q", got, nameA)
+	}
+	if !strings.Contains(nameA, "qf-studio/pilot") {
+		t.Errorf("expected tracker name to embed the repo, got %q", nameA)
+	}
+}
+
+// TestGithubHandlerSDK_LabelLifecycleTrackerPerRepo is a source-level guard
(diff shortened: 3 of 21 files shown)


alekspetrov:
## Post-merge review — PR#4867 (GH-4866, squash 2e5dfbae)

**Verdict: APPROVE** — all five spec legs land where the issue aimed them, and the central fix (the rule-set union) is the right shape: it repairs not just the deployed legacy 5-rule config but every future config that predates a later-added AlertType, at the single conversion both production callers funnel through. Three mutants killed, one survivor (D1 — an acceptance criterion explicitly asked for that mutant to die). One latent registration gap (D2), four notes. Nothing blocks the TASK-441 re-drill on the box's actual configuration.

### Spec-leg verification (isolated worktree @ 2e5dfbae)

1. **Rule-set union — correct layer, both callers covered.** The union lives inside `alerts.FromConfigAlerts` (config.go:32-58): build `seenTypes` from the caller's rules, append any `defaultRules()` entry whose Type is absent. Both production paths reach it — daemon `getAlertsConfig` now delegates to the new `config.AlertsConfig.ToAlertConfig` (config.go:467-517), and the gateway path's `initAlerts` ends in `FromConfigAlerts` directly (pilot.go:1722) — so the deployed legacy 5-rule config gets all four dead-man rules + the new `push_retry_exhausted_failure_streak` appended at load, no migration needed. The decision NOT to delegate `config.defaultAlertRules()` wholesale is justified in its comment and checks out: `failed_queue_high`/`api_error_rate_high`/`pr_stuck_waiting_ci` key off RuleCondition fields (`FailedQueueThreshold` etc.) that `AlertConditionConfig` cannot round-trip — full delegation would have zeroed those thresholds. The union appends the RICH `alerts.AlertRule` value directly, so no truncation on the unioned side either.
2. **`>=` + fired-once + loud crossing.** `RecordFailure` (deadman.go:216-258): `shouldFire := consecutive >= t.threshold && !t.fired` under the lock; the WARN (`alerts.deadman` component, tracker/alert_type/consecutive/threshold) fires BEFORE the `t.engine == nil` gate, so a crossing is greppable with no engine wired; `RecordSuccess` resets both the streak and the fired flag, so a seam that relapses after recovery pages again. All three caller-gated handlers (`handleDeadManStreak`, `handleIntentJudgeFailureStreak`, `handleDispatchLoopBreaker`) now track `matched` by Type regardless of Enabled and WARN on true no-match.
3. **Per-repo streak keying.** `labelLifecycleDeadManTrackerName(repoOwner+"/"+repoName)` at the event site (handlers.go:834) + startup registration per repo poller (poller_github.go:380), sharing counters via `RegisterDeadManTracker`'s memoize-by-name — one broken repo can no longer be zeroed by eleven healthy ones. Mirrors the intent-judge `intent_judge:`+repoFullName idiom. Both wirings are pinned by source-level guards (label_lifecycle_deadman_test.go), the established pattern for these otherwise-unexercisable cmd/pilot functions.
4. **Both seams wired, tested across the real boundary.** Label strip: attempt emitted right before the actual `ghEditLabels` call (after the legitimate skip checks — correct placement), failure with the underlying error / success on the outcome (lifecycle.go:444-486), via the same nil-safe `emitTripwireAlertEvent` relay the Finish sweep uses. Push-retry: one attempt per push *operation* (not per internal retry), failure only on exhaustion, success covers the RemoteBranchExists false-positive resolution (runner.go:4862-4921). Both seam tests drive a real `*ExecutionLifecycle`/`*Runner` into a recording `fakeAlertProcessor` via `SetAlertProcessor` and assert type+tracker+error payload — no argument-discarding mocks (TASK-441 L1 honored). The 403-shaped fake `gh` even reproduces the drill's archived-repo scenario.
5. **Doctor.** `checkAlertRuleCoverage` runs the config through the SAME conversion the daemon uses (`ToAlertConfig`) and diffs against `alerts.HandlerEmittedAlertTypes` via `CoverageGaps` — the legacy-5-rule test proves the union closes the gap, and the explicitly-disabled-rule case proves the check reads coverage, not just `Enabled`. The `checkAlertProcessorWiredOnRunner` flip is legitimate: main.go:3058 does call `runner.SetAlertProcessor` right after `alertsEngine.Start` (GH-4716), and pilot.go:1653 wires the orchestrator equivalent — the hardcoded `false` was a permanently-red doctor line, worse than no check.

Suites green at the merge commit: `internal/alerts` (1.7s) · `internal/health` (7.7s) · `internal/config` · `cmd/pilot` · both new executor seam tests.

### Mutation checks (code restored after each)

- **M1** — deleted the union loop in `FromConfigAlerts` → KILLED three packages deep: `TestFromConfigAlerts` + `TestFromConfigAlerts_Empty` (alerts), `TestCheckAlertRuleCoverage` (health), and the cmd/pilot acceptance table test. The load-bearing fix is pinned from every direction.
- **M2** — dropped `t.fired = false` from `RecordSuccess` → KILLED: `TestDeadManTracker_FiresAgainAfterRecoveryAndRelapse` (the relapse re-fire is real, not incidental).
- **M3** — deleted the `DeadManFailure` emit from the lifecycle strip-fail path → KILLED: the strip seam test fails on the missing failure event.
- **M4** — deleted the entire `if !matched { WARN }` block from `handleDeadManStreak` → **SURVIVES**: `go test ./internal/alerts/` stays green. See D1.

### Defects

**D1 — minor (test coverage) — internal/alerts/deadman_test.go:196 / engine_test.go — the no-match WARN is not asserted; the acceptance criterion's named mutant survives.** The issue's acceptance said: *"No-match path: with the rule deliberately removed, the new WARN is asserted (kills the silent-return mutation)."* The delivered tests (`TestHandleDeadManStreak_NoMatchingRule` + the two engine_test no-match subtests) assert only zero-alerts + no-panic — deleting the WARN block leaves the suite green (M4, proven). The injection point existed: `WithLogger` (engine.go:203) accepts any `*slog.Logger`, so a capturing handler could have pinned all three handler WARNs; deadman_test's "no logger injection point" comment is accurate only for the *tracker's* own `logging.WithComponent` WARN, not the engine handlers'. Impact is observability-only (a regression silently dropping the runtime backstop would ship green — the union and doctor check remain the load-bearing defenses), but this is the one acceptance line the PR explicitly doesn't meet. Follow-up: three `WithLogger`-captured assertions, ~30 LOC.

**D2 — minor (latent) — cmd/pilot/poller_github.go:397-419 — the two new global seam trackers are registered only inside the GitHub repo poller; every other daemon shape relays their events into a silent no-op.** `label_strip` and `push_retry_exhausted` registration lives solely in `startGithubSDKPollerForRepo`, but the seams that emit them run for EVERY adapter (the PR's own comment says so for push-retry). A daemon with no GitHub SDK repo poller — Telegram-only polling, or the gateway/orchestrator path (which wires the processor at pilot.go:1653 and registers only finish-tripwire trackers at :1665) — delivers the events to `namedDeadManTracker` misses, the documented silent no-op (deadman.go:49-51). No WARN fires either: the tracker is nil, not threshold-starved. The new doctor check can't see it — it reads *rule* coverage, not *tracker registration* — so it reports green on exactly this gap. Latent today (the box runs GitHub polling, so both trackers register), pre-existing in shape for `self_review` (same poller-only registration), but the PR extends the pattern to two new trackers while fixing this exact class one layer down. Follow-up shape: register the global (non-per-repo) seam trackers where the processor is wired (main.go:3058 vicinity + pilot.go initAlerts), and/or make a `namedDeadManTracker` miss log loudly — it is the last silent no-match left in this subsystem.

### Notes

- **N1**: `internal/pilot/pilot.go:1559-1722` still hand-rolls a third copy of the config→alerts conversion — `ToAlertConfig`'s "exactly one implementation" claim covers cmd/pilot + health only. The union still applies there (the copy ends in `FromConfigAlerts`), so no GH-4866-class risk, but Condition-field drift on the library path remains possible. Unification candidate for a later pass.
- **N2**: the acceptance's per-repo seam test ("10 failures for repo A with interleaved repo B successes still fires for A") was delivered as a name-function unit test + source-level grep guards rather than a behavioral interleaving test through the engine. Acceptable — per-name counter independence is structural (memoize-by-name, per-instance counters, both already pinned) — but the call-site wiring is grep-pinned, not behavior-pinned.
- **N3**: the `>=`-vs-`==` half of the fire-condition change is an equivalent mutant under the current API: counters start at 0 and threshold is fixed at registration, so `consecutive > threshold && !fired` is unreachable — the fired-once flag + reset-on-success (M2-pinned) is the real semantic change; the doc comment's "starts counting mid-streak, e.g. after a process restart" scenario can't occur with in-memory counters. Harmless defense-in-depth, worth knowing when reading the comment.
- **N4**: with `cfg.Alerts == nil` (nil engine), handlers.go's per-event registration still yields a fresh standalone tracker per issue (root-cause #3's second half), so label-lifecycle streaks can't accumulate to the greppable WARN on that path. Edge case — the poller-side registration covers every engine-present daemon — ledger only.

**Bottom line**: the class the kill-drill exposed — four rules that existed only on a code path the daemon never runs — is closed at the right chokepoint, mutation-verified from three packages, and live in v2.259.2. D1 and D2 are both real but neither touches the box's configuration; the re-drill (Drill A per the SOP) is now the meaningful test, and nothing found here predicts it failing.

Issue #4888, opened 2026-08-15
SendText has no thread parameter — progress chatter and answers escape the thread/topic (Slack affected today)


lkshrk:
### Summary

`comms.Messenger.SendText` has no `threadID` parameter while its three siblings (`SendConfirmation`, `SendResult`, `SendChunked`) do. The dozens of `SendText` call sites in `comms` — the progress chatter ("Looking into that...", "Drafting plan..."), the actual answers, cancellations and expiry notices — therefore cannot be threaded, regardless of adapter.

This is visible **today on Slack**: the Slack messenger threads confirmations/results via `ThreadTS`, but its `SendText` never sets `ThreadTS`, so a conversation started in a Slack thread gets its confirmations in-thread and its answers in the channel. For Telegram forum topics (#4871, #4887) the same gap sends everything to General. Commands lose the thread id one level earlier, in the telegram `handleCommand` wrapper.

The thread id is already available at every call site — the dispatcher passes `msg.ThreadID` into `handleQuestion` and friends; it is dropped at the `SendText` boundary.

### Steps to Reproduce

1. Slack adapter, start a conversation inside a thread (mention the bot in a thread reply).
2. Watch where the confirmation card vs. the progress messages and final answer land.

### Expected Behavior

Every reply piece lands in the thread (Slack) / topic (Telegram) the conversation is happening in.

### Actual Behavior

Confirmations/results thread correctly; everything sent via `SendText` lands at channel top level / General.

### Pilot Version

main @ 735c938f

### Adapter

Slack (affected today), Telegram (once forum topics land)

### Additional notes

Fix reaches further than the signature: `RunningTask` needs to carry `ThreadID` so cancellations answer in the right place, `cleanupExpiredTasks` should report expiry against the pending task's thread, and autopilot's Telegram notifier + brief sender currently hardcode the default destination. Have a working implementation covering telegram, slack, discord, web and the sdkshim bridge; will PR.

lkshrk:
Opened as #5121, stacked on #5120 (#4887). Reduces to a single SendText commit once #5120 merges.

Issue #4887, opened 2026-08-15
Telegram: replies should return to the forum topic a message arrived from


lkshrk:
### Problem

Follow-up to #4871. Outbound sends can target a configured forum topic, but the long-poll inbound path never captures which topic a message came from, so conversational replies fall back to the configured default no matter where the user was writing. A user asking a question in topic "Ops" gets the answer in whatever topic is configured — or General.

`comms.Handler` already forwards `msg.ThreadID` to every outbound messenger call — that is how Slack threads replies. The Telegram `Message` struct simply does not parse `message_thread_id`, and both `IncomingMessage` construction sites leave `ThreadID` empty. The SDK bridge path (`sdk_chat.go`) already maps it and works end to end for the same reason.

### Proposed Solution

Parse `message_thread_id` on `Message` and propagate it into `IncomingMessage.ThreadID`, gated on `is_topic_message`: Telegram also sets `message_thread_id` for ordinary reply chains in non-forum supergroups, but `sendMessage` accepts the parameter for forum supergroups only, so echoing that id back unconditionally would make the reply fail rather than thread.

Callback updates (approve/cancel buttons) carry the topic on `callback.Message` and should propagate it too, so a confirmation tapped inside a topic is answered there.

Have a working implementation, will open a PR once #4871 lands.

### Affected Adapters

Telegram

### Priority

Medium

alekspetrov:
Triage note on chain order: #4871's outbound leg merged tonight (PR #4891 — thanks for the careful `.agent/` cleanup and the parseMode revert). This issue is the natural next leg: inbound topic capture, gated on `is_topic_message` as you specced. Nothing on main parses `message_thread_id` from incoming updates yet, so the field is clear — please open the PR when ready and we'll review promptly. For the #4888 leg after it: the `SendText`/Messenger interface change touches all adapters, so keep any sdkshim bridge in the same PR so the build never has a half-threaded state. We're deliberately NOT dispatching our executor on this chain to avoid colliding with your in-flight branches.

alekspetrov:
[...]

PR #4891, opened 2026-08-15
feat(telegram): route notifications, approvals and alerts into a forum topic

### LINKED ISSUE #4871: feat: Telegram: support forum topics (message_thread_id) for notifications, approvals, and alerts
### Problem

Telegram supergroups with Topics enabled ("forums") route every message to a topic.
The Bot API targets one via the `message_thread_id` parameter on `sendMessage`.
Pilot can't set it, so task lifecycle notifications, approval prompts, and alerts
all land in the group's General thread.

Two things this blocks:

- **One group, several projects** — no way to give each project its own topic.
- **Critical vs routine alerts** — several `alerts.channels[]` entries can already
  differ by `severities`, but they all land in the same thread, so the split is
  invisible in the client.

There's no config-level workaround: `chat_id` can't encode a topic, and Pilot never
sends `reply_to_message_id`, which would otherwise thread a reply implicitly.

Notably the plumbing is already half-built. `internal/adapters/telegram/messenger.go`
accepts a `threadID` argument in three methods and uses it in none of them:

- `SendConfirmation(ctx, contextID, threadID, taskID, desc, project string)` — line 44
- `SendResult(ctx, contextID, threadID, taskID string, ...)` — line 81
- `SendChunked(ctx, contextID, threadID, content, prefix string)` — line 108

Each calls `m.client.SendMessage(...)` / `SendMessageWithKeyboard(...)` and discards
`threadID` — the code compiles fine while silently dropping the value.

The sibling adapters implement it: Slack sets `ThreadTS: threadID`
(`internal/adapters/slack/messenger.go:88`), Discord sets `ThreadID: ev.ThreadID`
(`internal/adapters/discord/handler.go:90`). Inbound already captures it t
`internal/adapters/telegram/sdk_chat.go:36` maps `ThreadID: ev.ThreadID` —
knows which topic a message arrived from and then answers in General.

Verified against `main` @ `2e85311f`.

### Proposed Solution

Additive and backwards compatible. `message_thread_id` is `omitempty`, so unset
serializes to nothing and existing non-forum configs are byte-identical on the wire.
(`omitempty` isn't optional here: sending `message_thread_id: 0` to a non-forum chat
makes the Bot API reject the call with `message thread not found`.)

- `SendMessageRequest` gains the field; `Client.SendMessage` /
  `SendMessageWithKeyboard` take a thread id. `EditMessage` and `AnswerCallback` are
  addressed by message/callback id and take no thread, so they're untouched — which
  also means pending-approval rehydration across restarts is unaffected.
- The messenger forwards the `threadID` it already receives.
- `adapters.telegram.message_thread_id` — default topic for lifecycle notifications
  and approval prompts.
- `alerts.channels[].telegram.message_thread_id` — per-channel topic, so several
  channels can share one `chat_id` and split by `severities`.

One design point worth agreeing on before review: approval prompts don't always go to
the configured chat. `resolveDestChatID` sends to `Approvers[0]` when that's a valid
Telegram destination — usually a private chat. Applying a topic id configured for the
group there would get `message thread not found` back and drop the prompt entirely,
which is worse than misplacing it. So the topic should apply only when the resolved
destination *is* the configured chat.

**Out of scope:** now that `sdk_chat.go` carries the inbound `ThreadID`, Pilot could
reply in the topic a message arrived from rather than a configured default. That
changes routing behaviour rather than adding a capability, so it belongs in its own
issue.

Happy to open the PR — I have a branch with the implementation, table-driven tests
(wire-level omit/emit, messenger forwarding including empty and unparseable thread
ids, alert-channel and approval-destination coverage), and docs updates for
`configs/pilot.example.yaml`, `docs/content/integrations/telegram.mdx`, and
`docs/content/features/alerts.mdx`.

### Alternatives Considered

- **A separate `chat_id` (separate group) per project or severity** — works today but
  fragments notifications across groups, multiplies setup, and loses the single-group
  overview that Topics exist to provide.
- **`reply_to_message_id`** — Pilot never sends it, and it threads to a *message*, not
  a topic, so it doesn't give configurable routing.
- **Parallel `...ToThread` client methods** instead of a parameter — avoids touching
  call sites, but duplicates every send path and leaves two ways to send a message.
- **Auto-reply in the inbound topic** — deferred as above; it changes existing routing
  rather than adding a capability.

### Affected Adapters

- [ ] GitHub
- [ ] GitLab
- [ ] Linear
- [ ] Jira
- [ ] Slack
- [ ] Discord
- [x] Telegram
- [ ] Azure DevOps
- [ ] Plane
- [ ] Asana

### Priority

Medium

lkshrk:
Adds `message_thread_id` support so notifications, approval prompts and alerts land in a configured forum topic instead of General. Non-forum chats are unaffected (`omitempty` guard — verified live).

Fixes #4871

Single commit; full rationale in the commit message. Verified: `go build/vet/test ./...` clean, `golangci-lint run --new-from-rev=main` 0 issues, running in production on a fork build.

Note: two follow-up PRs are stacked on this one (#4887 inbound-topic capture, #4888 SendText threading); I'll open each as its parent merges.

### DIFF
--- a/cmd/pilot/adapters.go
+++ b/cmd/pilot/adapters.go
@@ -29,7 +29,7 @@ type telegramBriefAdapter struct {
 }
 
 func (a *telegramBriefAdapter) SendBriefMessage(ctx context.Context, chatID, text, parseMode string) (*briefs.TelegramMessageResponse, error) {
-	resp, err := a.client.SendMessage(ctx, chatID, text, parseMode)
+	resp, err := a.client.SendMessage(ctx, chatID, text, parseMode, 0)
 	if err != nil {
 		return nil, err
 	}
@@ -44,8 +44,8 @@ type telegramApprovalAdapter struct {
 	client *telegram.Client
 }
 
-func (a *telegramApprovalAdapter) SendMessageWithKeyboard(ctx context.Context, chatID, text, parseMode string, keyboard [][]approval.InlineKeyboardButton) (*approval.MessageResponse, error) {
-	resp, err := a.client.SendMessageWithKeyboard(ctx, chatID, text, parseMode, convertKeyboardToTelegram(keyboard))
+func (a *telegramApprovalAdapter) SendMessageWithKeyboard(ctx context.Context, chatID, text, parseMode string, keyboard [][]approval.InlineKeyboardButton, messageThreadID int64) (*approval.MessageResponse, error) {
+	resp, err := a.client.SendMessageWithKeyboard(ctx, chatID, text, parseMode, convertKeyboardToTelegram(keyboard), messageThreadID)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/pilot/commands.go
+++ b/cmd/pilot/commands.go
@@ -688,7 +688,7 @@ Examples:
 					telegramClient := telegram.NewClient(cfg.Adapters.Telegram.BotToken)
 					for _, ch := range alertsCfg.Channels {
 						if ch.Type == "telegram" && ch.Telegram != nil {
-							telegramChannel := alerts.NewTelegramChannel(ch.Name, telegramClient, ch.Telegram.ChatID)
+							telegramChannel := alerts.NewTelegramChannel(ch.Name, telegramClient, ch.Telegram.ChatID, ch.Telegram.MessageThreadID)
 							dispatcher.RegisterChannel(telegramChannel)
 						}
 					}
--- a/cmd/pilot/main.go
+++ b/cmd/pilot/main.go
@@ -816,7 +816,7 @@ Examples:
 				if cfg.Adapters.Telegram != nil && cfg.Adapters.Telegram.Enabled && cfg.Adapters.Telegram.BotToken != "" &&
 					(cfg.Adapters.Telegram.Approval == nil || cfg.Adapters.Telegram.Approval.Enabled) {
 					tgClient := telegram.NewClient(cfg.Adapters.Telegram.BotToken)
-					gwTgApprovalHandler = approval.NewTelegramHandler(&telegramApprovalAdapter{client: tgClient}, cfg.Adapters.Telegram.ChatID)
+					gwTgApprovalHandler = approval.NewTelegramHandler(&telegramApprovalAdapter{client: tgClient}, cfg.Adapters.Telegram.ChatID, cfg.Adapters.Telegram.MessageThreadID)
 					// GH-3825: persist decisions directly to PRState via the manager so a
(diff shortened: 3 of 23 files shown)


alekspetrov (changes requested):
Thanks — the feature itself is solid: the `message_thread_id` plumbing is clean, the `omitempty` guard keeps non-forum chats byte-identical on the wire, the approver-override/topic interaction (DM gets topic 0, invalid-approver fallback keeps the topic) is handled correctly, and the test coverage — including the unset-topic fallback asserted at every layer — is above our bar. Requesting changes for one blocking issue and two nits:

**Blocking — the commit reverts repo agent-memory files.** It deletes `.agent/knowledge/memories/pitfalls/sequential-gates-on-execution-not-merge-fastfollow-misbase.md` and reverts `.agent/DEVELOPMENT-README.md` + `.agent/knowledge/graph.json` to a pre-08-15 snapshot (status/timestamp regressions plus unicode re-escaping). These look like a stale fork checkout swept in by `git commit -a`, not intentional — but they'd destroy project memory on merge. Please rebase on current `origin/main` (the PR is conflicting anyway) and drop all three `.agent/` paths from the commit; external PRs shouldn't touch `.agent/` at all.

**Nit 1:** `Notifier.SendMessage` quietly changed parseMode from `""` to `getParseMode()` — that flips plain-text sends to Markdown by default, which the Bot API can reject on arbitrary text ("can't parse entities"). Latent today, but out of scope and unremarked — please revert the hunk or split it out with rationale.

**Nit 2:** the `ThreadID` doc comment in `internal/comms/types.go` should note that for Telegram it now carries the forum topic id.

Ordering: we'll land #4892/#4893/#4894/#4907 first — they conflict textually with your call-site changes, and the rebase on your side is mechanical `, 0` additions, the same rebase that drops the `.agent/` files. Happy to approve immediately after.

alekspetrov (approved):
Re-review of 7d9e611b: all three items from the 08-17 review are resolved.

- **Blocker cleared**: the `.agent/` reversions are gone — the diff now touches only the feature surface (telegram/alerts/approval/comms plumbing + docs/config). Thanks for the careful rebase.
- **Nit 1 resolved**: plain-text `Notifier.SendMessage` keeps `parseMode ""`; the revision only threads `messageThreadID` through — no Markdown-by-default flip.
- **Nit 2 moot**: the new config-scoped `message_thread_id` design no longer touches `internal/comms/types.go`.

The feature verdict from the first review stands (clean `message_thread_id` plumbing, `omitempty` wire-compat, correct DM/topic interaction, layered fallback tests). CI green incl. the drift gate. Merging as promised.

PR #4892, opened 2026-08-15
fix(telegram): accept commands carrying the @botname suffix

### LINKED ISSUE #4880: Telegram: every command answers "Unknown command" in groups (@botname suffix not stripped)
### Summary

Telegram appends `@botname` to commands sent in groups, so `/status` arrives as `/status@yourbot`. `HandleCommand` switches on the raw token, which matches no case and answers "Unknown command" — every command is unusable in a group, the primary place the Telegram adapter runs.

`stripBotMention` already exists but only strips a *leading* mention (`@bot status`), not the suffix form Telegram actually sends.

### Steps to Reproduce

1. Add the bot to any group.
2. Send `/status` (Telegram clients auto-append `@yourbot` in groups).

### Expected Behavior

`/status@yourbot` behaves exactly like `/status`. A suffix naming a *different* bot returns without replying, so two bots in one group don't both answer the same command.

### Actual Behavior

"Unknown command" for every command sent in a group.

### Pilot Version

main @ 735c938f

### Adapter

Telegram

### Additional notes

Have a fix with table-driven tests including the other-bot case; will PR.

lkshrk:
Telegram appends `@botname` to commands in groups, so `/status` arrives as `/status@yourbot` and matched no case — every command answered "Unknown command" in groups. Strips the suffix for our own bot; a suffix naming a different bot returns without replying.

Fixes #4880

Single commit; full rationale in the commit message. Verified: build/vet/test clean, `--new-from-rev=main` 0 issues, running in production on a fork build.

### DIFF
--- a/internal/adapters/telegram/commands.go
+++ b/internal/adapters/telegram/commands.go
@@ -33,7 +33,10 @@ func (c *CommandHandler) HandleCommand(ctx context.Context, chatID, text string)
 		return
 	}
 
-	cmd := strings.ToLower(parts[0])
+	cmd, addressed := splitCommandMention(parts[0], c.handler.botUsername)
+	if !addressed {
+		return
+	}
 	args := parts[1:]
 
 	switch cmd {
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
Bug confirmed on main (`commands.go:36` lowercases the raw token, so `/status@bot` falls to Unknown command; `stripBotMention` only handles the leading-prefix form). Fix is correct and complete — `strings.Cut` + `EqualFold`, and the silent return for a suffix naming another bot is the right group etiquette. The fail-open when `getMe` failed is a reasonable availability tradeoff and it's documented in the test table. Test pins every branch.

PR #4893, opened 2026-08-15
fix(telegram): make the task confirmation buttons work

### LINKED ISSUE #4881: Telegram: Execute/Cancel confirmation buttons do nothing (callback payload mismatch)
### Summary

Tapping Execute or Cancel on a task confirmation does nothing. The card's buttons carry `execute_task:<id>` / `cancel_task:<id>` (set by `TelegramMessenger.SendConfirmation`), but `handleCallback` matches only the bare strings `"execute"` and `"cancel"` — which no producer in the repo emits. Both branches are unreachable; every tap falls through the switch.

It looks accepted rather than broken because `AnswerCallback` runs before the switch: Telegram clears the button's loading state, then nothing happens. No reply, no execution row; the confirmed task silently expires five minutes later.

### Steps to Reproduce

1. Send the bot any message that produces a task confirmation card.
2. Tap Execute.

### Expected Behavior

The task starts (or, for Cancel, is cancelled) and the user is told.

### Actual Behavior

Button spinner clears; nothing else happens; the pending task expires.

### Pilot Version

main @ 735c938f

### Adapter

Telegram

### Additional notes

Fix: match the prefixed payloads the buttons actually send, keeping the bare forms for the SDK bridge path that normalizes to them. Because this is a producer/consumer drift bug that unit tests on either side cannot catch, the fix includes a guard test asserting the payload prefixes emitted by `messenger.go` are matched in `handler.go`. Will PR.

lkshrk:
Buttons send `execute_task:<id>`/`cancel_task:<id>`; `handleCallback` matched only bare `execute`/`cancel` — both branches unreachable, every tap silently dropped. Now matches the prefixed payloads (bare forms kept for the SDK bridge). Includes a guard test asserting producer and consumer prefixes match so they can't drift again.

Fixes #4881

Single commit; verified: build/vet/test clean, `--new-from-rev=main` 0 issues, fix confirmed live (tap → task started → completed).

### DIFF
--- a/internal/adapters/telegram/callback_test.go
+++ b/internal/adapters/telegram/callback_test.go
@@ -1,7 +1,15 @@
 package telegram
 
 import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
 	"testing"
+
+	"github.com/qf-studio/pilot/internal/comms"
+	"github.com/qf-studio/pilot/internal/testutil"
 )
 
 // TestCallbackQueryData tests callback query data parsing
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
Confirmed dead-on: `messenger.go:49-50` emits `execute_task:<id>`/`cancel_task:<id>` while the switch matched only the bare strings — both branches unreachable, and `AnswerCallback` running first made it look accepted rather than broken. Keeping the bare forms is right (the sdkshim bridge messenger emits them).

The source-grep guard test is unusual but it's doing the real work here — it genuinely pins the producer/consumer prefix pair and fails on main. Approving.

One note for later, not this PR: the task ID in the payload is discarded (`ActionID` stays `"execute"`, and the comms handler confirms whatever is currently pending for the chat). That's the existing one-pending-per-context model, but if we ever allow multiple pending confirmations per chat, `handleConfirmation` needs to take the ID.

PR #4894, opened 2026-08-15
fix(telegram): surface messages dropped by allowed_ids instead of hiding them

### LINKED ISSUE #4882: Telegram: messages dropped by allowed_ids vanish silently (Debug-level log only)
### Summary

An inbound message failing the `allowed_ids` check logs at Debug and vanishes. With default log levels that is indistinguishable from the bot not receiving the message at all — no reply, no execution row, no log line.

The misconfiguration that triggers it is easy to make: `allowed_ids` listing only user ids while the group posts anonymously or as the group itself. In that case `From` is nil, the sender id is 0, and both allowlist lookups miss — every message from that group is eaten.

### Steps to Reproduce

1. Configure `allowed_ids` with user ids but not the group's chat id.
2. In the group, post as anonymous admin (or have the group itself post).

### Expected Behavior

A visible warning naming the chat id, the sender id, whether the message was anonymous, and which config key to edit.

### Actual Behavior

Nothing at Info level. The message is silently discarded.

### Pilot Version

main @ 735c938f

### Adapter

Telegram

### Additional notes

Proposed fix (implemented, will PR): warn instead of debug on the drop, dedupe warnings per chat/sender pair (following the `warnedInvalidDest` precedent in `internal/approval`), and warn at startup when `allowed_ids` is set but omits the configured `chat_id` — the case that silently drops every group message.

lkshrk:
Drops now warn (was Debug) naming chat id, sender id, anonymous flag and the config key to edit; warnings dedupe per chat/sender; startup warns when `allowed_ids` omits the configured `chat_id`.

Fixes #4882

Single commit; verified: build/vet/test clean, `--new-from-rev=main` 0 issues, running in production on a fork build.


alekspetrov (approved):
Real bug — Debug-level drops were invisible at production log level, and the anonymous-admin case (`sender_id=0`) was undiagnosable. Warn + named config key + the per-(chat,sender) dedupe is the right shape, and the startup `warnIfChatNotAllowed` cross-check catches the most common misconfiguration. Wiring is complete at both production constructor sites.

One ask, non-blocking: `TestWarnIfChatNotAllowed` currently asserts nothing (no slog capture), so it only proves no-panic. A capture handler asserting which cases warn would make it a real regression guard.

alekspetrov:
Merge conflict is the predicted trivial one — #4907 and this PR both append tests at the tail of handler_test.go, and #4907 landed first. Rebase on origin/main (keep both test blocks) and this merges; approval stands.

PR #4895, opened 2026-08-15
fix(linear): pass project_ids through to the SDK poller

### LINKED ISSUE #4883: Linear: project_ids is documented as a hard filter but never reaches the SDK poller (fails open)
### Summary

`adapters.linear.*.project_ids` is documented as a hard filter — "only issues belonging to those Linear projects are picked up. Issues without a project are skipped" — and has no effect. It parses, validates, and shows a healthy adapter in `pilot doctor`.

This fails **open**: an operator scoping Pilot to one project gets an agent that implements and opens PRs for issues they deliberately excluded. Reproduced on 2.259.2: two issues belonging to no project at all were polled, dispatched, and turned into PRs.

Root cause: the daemon polls Linear through the studio-sdk poller, and the config translation builds `linearSDK.WorkspaceConfig` from `Name`, `APIKey`, `TeamID`, `TriggerLabel` and `Polling` only — `ProjectIDs` is silently dropped at that literal. Everything on both sides already works: the SDK declares `ProjectIDs`, both poller list paths pass it into `ListIssues`, and `client.go` implements exactly the documented filter including skipping project-less issues.

### Steps to Reproduce

1. Configure `project_ids: [<one project>]`.
2. Label an issue outside that project (or with no project) with the trigger label.
3. Watch it get dispatched.

### Expected Behavior

Only issues in the listed projects are picked up; project-less issues are skipped.

### Actual Behavior

Everything with the trigger label is picked up.

### Pilot Version

main @ 735c938f (reproduced on 2.259.2)

### Adapter

Linear

### Additional notes

Fix implemented, will PR. Since the bug is a silently dropped field, the regression guard in the fix asserts the config translation itself (extracted into a testable function) rather than downstream behavior. Sibling pollers audited for the same class: github/gitlab/azuredevops omit `StaleLabelCleanup` (and github `Approval`/`App`) from their SDK literals, but all are consumed elsewhere in `cmd/pilot/main.go`; `project_ids` was the only field reaching no consumer at all.

lkshrk:
`project_ids` is documented as a hard filter and had no effect — dropped at the `linearSDK.WorkspaceConfig` translation literal, failing open. Translation moves into `newSDKLinearWorkspace` so the assignment itself is asserted by a regression guard (the bug class is a silently dropped field, so the guard belongs on the translation, not downstream behavior).

Fixes #4883

Single commit; full audit of sibling pollers in the commit message. Verified: build/vet/test clean, `--new-from-rev=main` 0 issues, filter confirmed live (in-project issue dispatched, project-less ignored).

### DIFF
--- a/cmd/pilot/poller_linear.go
+++ b/cmd/pilot/poller_linear.go
@@ -13,6 +13,21 @@ import (
 	"github.com/qf-studio/pilot/internal/logging"
 )
 
+func newSDKLinearWorkspace(name, apiKey, teamID, triggerLabel string, projectIDs, projects []string, interval time.Duration) *linearSDK.WorkspaceConfig {
+	return &linearSDK.WorkspaceConfig{
+		Name:         name,
+		APIKey:       apiKey,
+		TeamID:       teamID,
+		TriggerLabel: triggerLabel,
+		ProjectIDs:   projectIDs,
+		Projects:     projects,
+		Polling: &linearSDK.PollingConfig{
+			Enabled:  true,
+			Interval: interval,
+		},
+	}
+}
+
 func linearPollerRegistration() PollerRegistration {
 	return PollerRegistration{
 		Name: "linear",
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
Approve. Bug confirmed at `poller_linear.go:47-56` — the `WorkspaceConfig` literal silently drops the documented hard filter; verified the SDK consumes it down to the `ListIssues` client-side filter, so the fix is effective. Unfiltered path unchanged and covered.

Optional: the regression guard tests the translation helper, but nothing stops a future call site reverting to an inline literal — this file already has source-grep call-site tests, one more would fully close the class.

Issue #4889, opened 2026-08-15
Linear: report unusable labels at startup instead of degrading silently


lkshrk:
### Problem

Two Linear label problems are each diagnosable only by their symptoms.

A **workspace-scoped trigger label** is visible in Linear's UI and appears under the team, but the poller filters `issueLabels` by team key and never matches it — the daemon dies with "label X not found in team ROU" while the operator is looking straight at the label. Linear creates labels workspace-scoped by default from the issue view, and scope is immutable, so recovery is delete, recreate team-scoped, re-apply — none of which the error suggests.

The **`pilot-*` status labels** fail differently (see #4884): creation failures are logged at Warn deep in the poller and execution continues, so issues run but are never marked in-progress/done/failed — silent loss of the whole status workflow.

### Proposed Solution

A startup preflight that classifies each configured label as team-scoped, workspace-scoped, another team's, or missing, and logs the specific remedy for each. Behavior otherwise unchanged: the trigger label still fails closed, status labels still degrade — but neither is silent about why.

Implemented, will PR after the team-key fix (stacked on it).

### Alternatives Considered

Fixing only the error messages at the point of failure — but the failures surface far from their cause and some (status labels) are mid-poll where an operator isn't looking. A single startup report is where misconfiguration is actually noticed.

### Affected Adapters

Linear

### Priority

Medium

---

## Maintainer addendum (2026-08-21, pre-dispatch)

**Anchor update**: the production Linear poller runs the **studio-sdk client** (`cmd/pilot/poller_linear.go` → `studio-sdk/sdk/integrations/linear`), NOT `internal/adapters/linear` — write the preflight against the SDK client or it validates a path production doesn't run. Since this issue was filed, sdk PR#115 (workspace-label fallback) + PR#116 (team key→UUID) merged — once the pin bump lands, the workspace-scoped startup DEATH disappears; this preflight becomes the loud-diagnostics layer for the remaining modes (other-team labels, missing labels, status-label creation failures).

## Acceptance Criteria

- [ ] At startup, for each configured Linear workspace: classify the trigger label and each `pilot-*` status label as team-scoped / workspace-scoped / another-team's / missing, and log one line per problem with the specific remedy (the classification query goes through the sdk client the poller uses).
- [ ] Behavior otherwise unchanged: trigger label still fails closed; status labels still degrade — but never silently (each degradation names its cause).
- [ ] Preflight failure of the CHECK itself (API error) is WARN + continue — never blocks startup.
[...]

PR #4896, opened 2026-08-15
fix(linear): resolve the team key to a UUID before creating labels

### LINKED ISSUE #4884: Linear: label lookup wants the team KEY, label creation wants the UUID — one of the two always breaks
### Summary

`GetLabelByName` filters `team: { key: { eq: $teamId } }` and needs the team **key** (e.g. `ROU`), while `CreateLabel` passes the same configured value straight into `issueLabelCreate(input: { teamId: ... })`, which needs the team **UUID**. Whichever form `team_id` holds, one path fails: with a key, label creation fails; with a UUID, issue and label lookup silently match nothing.

This is why every `pilot-*` status label has to be pre-created by hand today.

(Related: the mutation contract violation also exists inside studio-sdk itself — filed as qf-studio/studio-sdk#113. This issue is about pilot's own legacy-client path.)

### Steps to Reproduce

1. Configure `team_id` with the documented team key.
2. Start the daemon in a workspace without pre-created `pilot-*` labels.

### Expected Behavior

The documented team key works for lookup and creation alike; status labels are auto-created.

### Actual Behavior

Label creation fails (logged at Warn deep in the poller, execution continues) — issues run but are never marked in-progress/done/failed.

### Pilot Version

main @ 735c938f

### Adapter

Linear

### Additional notes

Fix implemented, will PR: `CreateLabel` resolves a key to its UUID first, leaving a configured UUID untouched.

lkshrk:
`GetLabelByName` needs the team KEY, `CreateLabel` needs the UUID — whichever form `team_id` holds, one path breaks (why pilot-* labels must be pre-created by hand). `CreateLabel` now resolves a key to its UUID first; a configured UUID passes through untouched.

Fixes #4884

Related SDK-side contract violation filed as qf-studio/studio-sdk#113. Follow-up PR stacked on this one adds startup diagnostics (#4889).

Single commit; verified: build/vet/test clean, `--new-from-rev=main` 0 issues.

### DIFF
--- a/internal/adapters/linear/client.go
+++ b/internal/adapters/linear/client.go
@@ -435,10 +435,92 @@ func (c *Client) GetLabelByName(ctx context.Context, teamID, labelName string) (
 	}
 
 	if len(result.IssueLabels.Nodes) == 0 {
-		return "", fmt.Errorf("label %q not found in team %s", labelName, teamID)
+		return c.getWorkspaceLabelByName(ctx, teamID, labelName)
 	}
 
 	return result.IssueLabels.Nodes[0].ID, nil
+}
+
+func (c *Client) getWorkspaceLabelByName(ctx context.Context, teamID, labelName string) (string, error) {
+	query := `
+		query GetWorkspaceLabel($name: String!) {
+			issueLabels(filter: { name: { eq: $name } }) {
+				nodes { id name team { id } }
+			}
+		}
+	`
+	var result struct {
+		IssueLabels struct {
+			Nodes []struct {
+				ID   string `json:"id"`
+				Name string `json:"name"`
+				Team *struct {
+					ID string `json:"id"`
+				} `json:"team"`
+			} `json:"nodes"`
+		} `json:"issueLabels"`
+	}
+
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
Approve with two notes. The key→UUID resolution before `issueLabelCreate` is exactly right, `looksLikeUUID` passthrough means UUID-configured deployments see zero extra requests, and the test matrix (resolve-vs-passthrough, workspace fallback, another-team's-label rejection, get-or-create) is thorough.

1. #4884's other half is still open: `GetLabelByName` filters by `team.key` only, so a UUID-configured `team_id` never finds team-scoped labels and `GetOrCreateLabel` re-attempts the create every startup. The symmetric fix is filtering `team: { id: { eq } }` when `looksLikeUUID(teamID)` — happy to take that as a follow-up on our side if you'd rather keep this PR scoped, just say which.
2. The workspace-label fallback is beyond the issue's scope — it's coherent and well-tested, keep it, but a sentence in the PR body acknowledging it would help future archaeology.

PR #4897, opened 2026-08-15
fix(github): verify App installation tokens without calling /user

### LINKED ISSUE #4885: GitHub App installation tokens always fail verification (Verify calls /user)
### Summary

`Verify` calls `GetAuthenticatedUser`, i.e. `GET /user`. That endpoint is user-scoped and a GitHub App installation token structurally cannot call it, so every App-auth deployment has its GitHub adapter marked as failing at startup regardless of whether the credential is good.

The practical effect is worse than a bad status line: preflight emits `EventTypeConfigError` on that failure, so App auth produces a **permanent `service_unhealthy` alert**, and the documented workaround is to leave `adapters.github.app` unset and get bot authorship from a `gh` shim instead.

### Steps to Reproduce

1. Configure `adapters.github.app` with a valid App credential.
2. Start the daemon; watch preflight.

### Expected Behavior

A valid installation token verifies clean.

### Actual Behavior

Verification fails; `service_unhealthy` alert fires and never resolves.

### Pilot Version

main @ 735c938f

### Adapter

GitHub

### Additional notes

Fix implemented, will PR: a `/user` failure is treated as inconclusive rather than fatal — the credential is re-probed against `GET /installation/repositories`, valid for installation tokens; only failure of both reports an invalid token. User and classic tokens still verify on the first call and never reach the second, so their behavior and request count are unchanged.

lkshrk:
`Verify` called `GET /user`, which installation tokens structurally cannot call — every App-auth deployment failed verification and carried a permanent service_unhealthy alert. A /user failure is now inconclusive: re-probe against `GET /installation/repositories`; only failure of both reports an invalid token. User/classic tokens verify on the first call, unchanged behavior and request count.

Fixes #4885

Single commit; verified: build/vet/test clean, `--new-from-rev=main` 0 issues, App-auth deployment verifies clean in production on a fork build.

### DIFF
--- a/internal/adapters/github/client.go
+++ b/internal/adapters/github/client.go
@@ -1275,13 +1275,24 @@ func (c *Client) GetAuthenticatedUser(ctx context.Context) (*User, error) {
 // dead token can be diagnosed without re-deriving the resolution chain
 // (GH-3718). Pass "" when the source is unknown.
 func (c *Client) Verify(ctx context.Context, tokenSource string) error {
-	if _, err := c.GetAuthenticatedUser(ctx); err != nil {
-		if tokenSource != "" {
-			return fmt.Errorf("github token invalid (source: %s): %w", tokenSource, err)
-		}
-		return fmt.Errorf("github token invalid: %w", err)
+	_, err := c.GetAuthenticatedUser(ctx)
+	if err == nil {
+		return nil
 	}
-	return nil
+	if instErr := c.verifyInstallation(ctx); instErr == nil {
+		return nil
+	}
+	if tokenSource != "" {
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
Approve. `GET /installation/repositories?per_page=1` is the canonical installation-token self-probe; a dead PAT fails both probes so no false positives, a valid installation token with zero repos correctly passes, and the PAT path stays request-count-identical (pinned by the `wantInstallProbe: false` row). Probe rides the existing `doRequest` — no credential exposure, and a /user 403 correctly doesn't trip the AuthError invalidation path.

Nit: on double failure only the /user error is returned; for App deployments the installation-probe error is the diagnostic one — consider `%w (installation probe: %v)`.

Issue #4890, opened 2026-08-15
Alerts: persist active-alert state so recovery survives a restart


lkshrk:
### Problem

Follow-up to #4886. Active alerts live only in the engine's map, so a condition that recovers while the daemon is down produces no resolution at all: the restarted process has no record anything fired, and the healthy event is correctly but unhelpfully silent. Restarting is the most likely thing an operator does right after fixing a credential — exactly when recovery reporting is least able to work.

### Proposed Solution

Follow the pending-approval precedent: a new `active_alerts` table, an optional `ActiveAlertStore` the engine writes through, and a rehydrate call on startup. `*memory.Store` satisfies the interface directly.

Only the fields needed to emit a resolution are stored — including the channels the original alert was delivered to, so a rehydrated resolution still reaches those channels rather than being re-filtered on its own info severity.

Persistence is best-effort and off the alerting path: a store failure is logged and the alert still fires, so a broken database degrades recovery reporting rather than suppressing alerts. The store is optional; engines constructed without one behave exactly as before.

Implemented and running in production, will PR after resolution-notifications (stacked on it).

### Alternatives Considered

Re-deriving active state from the alert history table — history has no notion of "still firing" and doesn't record delivery channels.

### Affected Adapters

All (alerting core)

### Priority

Medium

---

## Maintainer addendum (2026-08-21, pre-dispatch)

Stacking blocker resolved: #4886 closed via PR#4898 (merged 08-17). Design as proposed is accepted — pending-approval persistence precedent, optional store, best-effort writes.

## Acceptance Criteria

- [ ] `ActiveAlertStore` optional interface in `internal/alerts`; `*memory.Store` satisfies it (compile-time assertion). Engines constructed without a store behave byte-identically (regression-pinned).
- [ ] New `active_alerts` table via the established duplicate-tolerant migration idiom (`internal/memory`); rows store only what a resolution needs, INCLUDING the channels the original alert was delivered to.
- [ ] Writes are best-effort and off the alerting path: store failure → log + alert still fires (test with a failing store).
- [ ] Rehydrate on startup; a condition that recovered while the daemon was down emits its resolution to the ORIGINALLY-delivered channels, not re-filtered by severity (real-SQLite e2e test: fire → persist → new engine instance → rehydrate → resolve → assert channel set).
- [ ] Firing/resolution lifecycle keeps the table bounded (resolved rows deleted or marked; no unbounded growth — state the choice).
[...]

PR #4898, opened 2026-08-15
feat(alerts): notify when an alerted condition recovers

### LINKED ISSUE #4886: Alerts: notify when an alerted condition recovers
### Problem

Alerts fire on degradation and never report recovery. An operator sees `service_unhealthy: github adapter verification failed ...` and then silence, with no way to distinguish "fixed a minute later" from "still broken a week on" short of going and checking. One-directional alerts get muted, which defeats the subsystem.

The data model already expects this: `Alert.ResolvedAt` is declared with no production writer anywhere in the repo. The recovery signal is already computed and thrown away — `runAdapterPreflight` logs "adapter verified" on success and emits nothing, while only the failure branch emits `EventTypeConfigError`.

### Proposed Solution

- Engine tracks active alerts keyed by rule + source, alongside the existing per-rule state.
- Adapter preflight emits `EventTypeConfigHealthy`; both outcomes carry a stable `adapter:<name>` source so multiple adapters resolve independently.
- Recovery sets `ResolvedAt`, dispatches a RESOLVED notification carrying how long the condition lasted, clears the active entry so a regression alerts again.
- `alerts.defaults.notify_on_resolve` (pointer, unset = enabled) gates it.
- `alerts.health_check_interval` (default 15m, negative disables) re-runs the verifiers so a credential fixed at 10:00 resolves without waiting for the next restart.

Two deliberate choices worth review: resolutions **bypass the cooldown** (cooldown suppresses repeat-firing of a persistent problem; suppressing a recovery because the failure was recent is the opposite of what an operator wants), and resolutions **route to the channels that received the original alert** rather than re-filtering on the resolution's own info severity — a channel configured `severities: [warning, critical]` would otherwise receive the failure and silently drop the recovery.

Scope is credential/adapter health. `task_failed` and `github_sideeffect` are events rather than states and are deliberately untouched; `task_stuck`, `consecutive_failures`, `daily_spend_exceeded`, `budget_depleted` are state-shaped natural follow-ups.

Implemented and running in production, will PR.

### Alternatives Considered

Startup-only resolution (no health-check interval) — would rarely fire, since the condition usually recovers while the daemon is up.

### Affected Adapters

All (alerting core)

### Priority

Medium

lkshrk:
Alerts fired on degradation and never reported recovery; `Alert.ResolvedAt` existed with no production writer. Engine now tracks active alerts by rule+source, preflight emits `EventTypeConfigHealthy`, recovery dispatches a RESOLVED notification with duration. Gated by `alerts.defaults.notify_on_resolve` (unset = enabled); `alerts.health_check_interval` (default 15m) re-runs verifiers so recovery doesn't wait for a restart.

Two deliberate choices called out for review in the commit message: resolutions bypass cooldown, and route to the channels that received the original alert (not re-filtered on info severity).

Fixes #4886

Follow-up PR stacked on this one persists active state across restarts (#4890).

Single commit; verified: build/vet/test clean, `--new-from-rev=main` 0 issues, running in production on a fork build.

### DIFF
--- a/cmd/pilot/adapter_preflight.go
+++ b/cmd/pilot/adapter_preflight.go
@@ -76,6 +76,10 @@ func buildAdapterVerifiers(cfg *config.Config) []verify.Verifiable {
 	return verifiers
 }
 
+func adapterAlertSource(name string) string {
+	return "adapter:" + name
+}
+
 // runAdapterPreflight calls Verify(ctx) on every verifier concurrently,
 // each bounded by preflightVerifyTimeout, and returns the per-adapter
 // result. A failure logs ERROR and emits an alerts.EventTypeConfigError
@@ -110,6 +114,13 @@ func runAdapterPreflight(ctx context.Context, verifiers []verify.Verifiable, ale
 			switch {
 			case err == nil:
 				log.Info("adapter verified", slog.String("adapter", v.Name()))
+				if alertsEngine != nil {
+					alertsEngine.ProcessEvent(alerts.Event{
+						Type:      alerts.EventTypeConfigHealthy,
+						Source:    adapterAlertSource(v.Name()),
+						Timestamp: time.Now(),
+					})
+				}
 			case errors.Is(err, verify.ErrProbeNotImplemented):
 				// No live probe yet — not evidence of a broken credential.
 			default:
@@ -120,6 +131,7 @@ func runAdapterPreflight(ctx context.Context, verifiers []verify.Verifiable, ale
 				if alertsEngine != nil {
 					alertsEngine.ProcessEvent(alerts.Event{
 						Type:      alerts.EventTypeConfigError,
+						Source:    adapterAlertSource(v.Name()),
 						Error:     fmt.Sprintf("%s adapter verification failed: %v", v.Name(), err),
 						Timestamp: time.Now(),
 					})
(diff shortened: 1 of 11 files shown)


alekspetrov (approved):
Solid feature — the engine state model, per-source keying, and the Dispatch-with-explicit-channels severity bypass all check out (verified the info-severity resolution genuinely reaches warning-only channels), and the deliberate calls (cooldown bypass for resolutions, original-channel routing) are the right ones. Approving with three asks before merge:

1. Add `notify_on_resolve` and `alerts.health_check_interval` to `configs/pilot.example.yaml` — it's the full config reference, and "negative disables, 0 = 15m default" is unguessable from YAML.
2. Flap-under-cooldown: after a resolve, a re-fire inside the rule cooldown is suppressed by `shouldFire`, leaving "✅ Resolved" as the last message while the condition is broken again for up to cooldown + one tick. Either exempt the first post-resolution re-fire from cooldown, or add a test + comment documenting the window (the current regression test only exercises cooldown=0).
3. `dispatchResolution` calls `RecordFired`, inflating the fired-alerts counter — consider a distinct metric or label for resolutions.

Restart persistence deferred to #4890 is fine — please note there that it depends on this PR's `activeAlerts` shape.

Issue #2777, opened 2026-05-07
feat(executor): DECLINED-with-reason structured output for unactionable tasks


alekspetrov:
Parent: GH-2753

Implement the full DECLINED path end-to-end as a single unit per the task's explicit "keep as single issue" directive. Add `parseDeclinedReason` helper and `ErrorTypeDeclined` constant in `internal/executor/runner.go`, rewrite the retry prompt at lines 2176-2188 to offer implement-or-DECLINED outcomes with examples, branch on parse at lines 2228-2253 to set `result.Declined`/`result.DeclinedReason` and status `declined` without failure escalation, extend the Result type with the new fields, wire GitHub-side label flow (`pilot-needs-clarification` add, `pilot-in-progress` remove, comment with reason + clarification hint) in the executor's GitHub integration path, add `pilot-needs-clarification` to the blocking-label set in `internal/adapters/github/poller.go`, surface a `declined` count in the dashboard alongside completed/failed, verify the `executions.status` schema accepts `declined` (extending the allow-list if needed), add table-driven tests for `parseDeclinedReason`, `TestRunner_DECLINED_Path`, `TestRunner_NoChanges_NoDecline`, and a poller skip-on-label test, document the new behavior in `docs/content/features/...`, and ensure `make test`, `make lint`, `make build` pass. Splitting this would fragment runner.go edits across parallel subtasks and risks the cascade patterns flagged in the linked bug notes; the entire change is ~150 LoC and stays cohesive as one PR.

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-2777 is already queued or running
```

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-2777 is already queued or running
```

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-2777 is already queued or running
```

alekspetrov:
❌ Pilot execution failed:

```
failed to queue task: task GH-2777 is already queued or running
```

alekspetrov:
❌ Pilot execution failed:

```
execution failed: subtask 2/3 failed: oom_killed: Process killed by SIGKILL (exit code 137)
```

alekspetrov:
✅ Pilot completed!

| Metric | Value |
|--------|-------|
| Duration | 17m48s |
| Branch | `pilot/GH-2777` |
[...]

Issue #4517, opened 2026-07-23
fix(executor): dirty worktree at session end is deleted as no_op — 44 min of completed, test-passing B8 work destroyed


alekspetrov:
## Problem

When an execution ends with NO new commit, the executor classifies `no_op` ("no new commit produced — worktree HEAD matches base branch parent") and the isolated worktree is deleted — **including any uncommitted work in it**.

Live incident (2026-07-23, `pilot-console#26` / B8, recording `TG-1784822448733`): a 43m54s run produced 191,535 output tokens across multiple Implementing/Testing phases; its final message claimed all 7 ACs complete and cited the full `internal/fleet` suite passing — which it did, IN THE WORKTREE, because the implementation existed there as uncommitted files. The session never ran `git commit`. Harvester: `CommitSHA recovered via git sha=c172845` (= base main HEAD) → `no_op` → worktree `/tmp/pilot-worktree-GH-26-1784822447012758019` deleted. ~44 minutes of completed, test-passing work destroyed. ~$2.87 spent.

Contributing model behavior: after context compaction mid-session, the model attributed its own pre-compact work to "earlier sessions" ("AC1–AC5 done in earlier sessions; AC6 verified in this session") and went straight to verification, skipping the commit step. The executor has no backstop for this.

## Fix

Two layers:

1. **Harvester backstop (the real fix)**: before classifying `no_op`, run `git status --porcelain` in the worktree. If dirty (untracked/modified files beyond noise like `.agent/` scratch), do NOT delete silently:
   - Preferred: auto-commit the dirty state to the task branch as `wip(<task-id>): uncommitted session work (auto-preserved)` and push the branch, then classify as failed-with-artifact (NOT completed) so retry/review can pick it up — never lose the work.
   - Minimum: preserve the worktree (move to a quarantine dir with a TTL) and log its path in the execution row error field.
2. **Prompt hardening**: the execution prompt's completion contract should state explicitly that work is only complete after `git commit` on the task branch, and that after any context compaction the model MUST verify `git log`/`git status` reflects its work before concluding — "tests pass" in a dirty tree is not done.

## Acceptance criteria

- AC1: table-driven test — session ends, worktree dirty (new + modified files), harvester finds no commit → dirty state auto-committed to the task branch and pushed; execution NOT classified `no_op`; worktree cleanup only after the push succeeds.
- AC2: clean-tree no-commit case still classifies `no_op` exactly as today (genuine no-ops unchanged).
[...]

Issue #3126, opened 2026-05-26
TASK-300-redo: Executor ghost-SHA guard — test fixture must create main branch first


alekspetrov:
## Context

Redo of #3112 (ghost-closed despite CI failure). The new ghost-SHA guard (TASK-300) needs a corrected test fixture — a fresh `git init` repo has no `main` branch ref, so the existing fixture's `push -u origin main` step fails before the assertion runs. This issue ships the fixture fix plus the full TASK-300 scope below.

## Acceptance

- `go test ./internal/executor/... -run TestCommitSHAIsNew -count=1 -v` passes locally
- Pre-PR verification commands in the section below all pass
- Files listed in "## Files" section exist and compile

---

## Redo of #3112 — ghost-closed despite CI failure

**Original:** #3112 / PR #3116. CI test failed (legitimate test-fixture bug, not infra flake). PR was closed without merge AND issue was marked `pilot-done` (Variant 5 ghost-close pattern). Re-opening #3112 did not trigger Pilot to re-execute (processed-cache holds the original issue number). Fresh issue # to bypass.

## What broke last time

`internal/executor/git_freshness_test.go` test fixture does:
```go
tmpDir := t.TempDir()
runGit(t, tmpDir, "init")
// ...
runGit(t, tmpDir, "push", "-u", "origin", "main")  // ← FAILS: src refspec main does not match any
```

A freshly `git init`-ed repo has no `main` branch — there's no initial commit, no branch reference. The push attempt fails with `error: src refspec main does not match any`. Two tests fail: `TestCommitSHAIsNew_SHAOnMain`, `TestCommitSHAIsNew_SHAOnFeatureBranch`.

## Fix pattern (apply this in the redo)

```go
tmpDir := t.TempDir()
runGit(t, tmpDir, "init", "-b", "main")          // create with main as initial branch (modern git)
// OR:
//   runGit(t, tmpDir, "init")
//   runGit(t, tmpDir, "checkout", "-b", "main")
runGit(t, tmpDir, "commit", "--allow-empty", "-m", "initial")  // CREATE a commit so main branch reference exists

// For tests that need an "origin" remote:
bareDir := t.TempDir()
runGit(t, bareDir, "init", "--bare")
runGit(t, tmpDir, "remote", "add", "origin", bareDir)
runGit(t, tmpDir, "push", "-u", "origin", "main")  // NOW this succeeds
```

[...]

Issue #916, opened 2026-02-12
feat(executor): detect and retry when Claude makes no commits


alekspetrov:
## Context

~10% of failures are "No commits between main and pilot/GH-XXX". Claude Code runs successfully but makes no actual changes, then PR creation fails.

## Task

In `internal/executor/runner.go`, after backend execution completes but before PR creation (~line 900), add:

```go
// Check if any commits were made
git := NewGitOperations(task.ProjectPath)
commitCount, err := git.CommitCountSinceMain(ctx)
if err != nil {
    log.Warn("Failed to count commits", slog.Any("error", err))
}

if commitCount == 0 && !task.DirectCommit {
    // No commits made - this will fail PR creation
    // Retry once with explicit instruction
    log.Warn("Claude made no commits, retrying with explicit instruction",
        slog.String("task_id", task.ID),
    )
    
    retryPrompt := fmt.Sprintf(`## Retry: No Changes Detected

The previous execution completed but made no code changes. This task requires actual implementation.

**Original Task:** %s

**Instructions:**
1. Read the task requirements carefully
2. Implement the required changes
3. Create at least one commit with your changes
4. Do NOT just analyze or plan - actually write and commit code

%s`, task.Title, task.Description)

    // Retry once
    retryResult, retryErr := r.backend.Execute(ctx, ExecuteOptions{
        Prompt:      retryPrompt,
        ProjectPath: task.ProjectPath,
        // ... other options
    })
    
    // Check again after retry
    commitCount, _ = git.CommitCountSinceMain(ctx)
    if commitCount == 0 {
        return &ExecutionResult{
            Success: false,
            Error:   "Claude completed but made no code changes after retry",
        }, nil
    }
}
```

Add to `internal/executor/git.go`:

```go
func (g *GitOperations) CommitCountSinceMain(ctx context.Context) (int, error) {
    cmd := exec.CommandContext(ctx, "git", "rev-list", "--count", "origin/main..HEAD")
    cmd.Dir = g.projectPath
    out, err := cmd.Output()
    if err != nil {
        return 0, err
    }
    count, _ := strconv.Atoi(strings.TrimSpace(string(out)))
    return count, nil
}
```

## Acceptance Criteria

- [ ] `CommitCountSinceMain()` added to git.go
- [ ] No-commits detection in runner.go after Claude execution
- [ ] One retry with explicit instruction prompt
- [ ] Clear error if still no commits after retry
- [ ] Tests for the detection logic
[...]

PR #4901, opened 2026-08-15
fix(executor): honor a first-pass decline and render declines as no-ops

### LINKED ISSUE #4875: Correct executor decline rendered as ❌ Task failed; no-commits retry fires before decline is recognized
### Summary

A correct executor decline (no-op with rationale) is rendered to the user as ❌ Task failed, and the no-commits retry fires before the decline is recognized, burning a second model call to re-attempt work the first call already correctly declined.

A correct no-op that looks identical to a crash trains operators to ignore ❌. Watch-only and already-satisfied tickets are common enough that this recurs.

### Steps to Reproduce

1. Give the executor a task with no code change to make (here: a watch-only ticket tracking an external upstream issue whose trigger condition is unmet).
2. Executor completes, reasons correctly, declines with an explicit rationale.

### Expected Behavior

`declined` rendered distinctly from `failed` (e.g. ⏸ / "no action needed" plus the reason, which is already captured in the decline log). ❌ reserved for actual failures. The decline checked *before* the no-commits retry, so a reasoned no-op is not re-invoked.

### Actual Behavior

Telegram showed:

```
🚀 Starting TG-1786817399...
❌ Task failed: TG-1786817399
<followed by a correct, complete analysis and an explicit NO-OP RATIONALE>
```

And in sequence: task completes cleanly (`is_error=false`), then `Claude made no commits, retrying with explicit instruction` (`runner.go:3989`) fires, and only after the retry does `Task declined by executor` (`runner.go:4089`) surface — the retry trigger sits upstream of decline detection.

### Pilot Version

v2.260.3 (fork build of main @ 735c938f; executor paths unmodified from upstream)

### Adapter

Telegram (rendering; the sequencing bug is adapter-independent)

### Logs

```
18:13:12 DEBUG Backend result received component=executor task_id=TG-1786817399 is_error=false
18:13:12 INFO  Task completed component=executor task_id=TG-1786817399 complexity=simple
               duration=1m36.177466455s tokens_in=545 tokens_out=3834 cost_usd=0.6401928
               files_read=2 files_written=0
18:13:14 WARN  Claude made no commits, retrying with explicit instruction
               component=executor task_id=TG-1786817399 branch=pilot/TG-1786817399
18:13:37 WARN  Task declined by executor component=executor task_id=TG-1786817399
               reason="ROU-448 is a watch-only ticket tracking an external upstream GitHub
               issue; it was already re-checked today and logged as unchanged — there is
               no code change to make"
```

lkshrk:
A correctly declined task rendered as `❌ Task failed` — indistinguishable from a crash — and the no-commit retry fired before the decline was recognized, burning a second model call to re-attempt work the first call had just explained it should not do.

Two connected gaps, both fixed:

1. **Retry-before-decline**: the `DECLINED:<reason>` escape hatch (GH-2777) was only checked *after* the no-commit retry. The first execution's `LastAssistantText` is now checked for the marker before retrying; decline handling is shared via a closure so both sites behave identically. A model that declines up front costs one call instead of two.
2. **Decline dies at the messenger boundary**: comms passed only `result.Success` to `SendResult`, so `Declined` was erased and every messenger rendered a decline as a failure. The handler now short-circuits declined results into a distinct `⏸ No changes needed` message carrying the captured reason; ❌ stays reserved for actual failures.

Guard test drives a declining backend through `Execute` and asserts exactly one backend invocation and a declined outcome.

Fixes #4875

Verified: comms + executor suites green, `golangci-lint --new-from-rev=main` 0 issues. Live incident evidence in the issue (watch-only ticket, correct no-op rationale, rendered ❌ after a wasted retry).

### DIFF
--- a/internal/comms/handler.go
+++ b/internal/comms/handler.go
@@ -844,6 +844,15 @@ func (h *Handler) executeTaskCore(ctx context.Context, contextID, threadID, task
 		return
 	}
 
+	if result.Declined {
+		reason := result.DeclinedReason
+		if reason == "" {
+			reason = "no action needed"
+		}
+		_ = h.messenger.SendText(ctx, contextID, fmt.Sprintf("⏸ No changes needed: %s\n%s", taskID, reason))
+		return
+	}
+
 	output := CleanInternalSignals(result.Output)
 	_ = h.messenger.SendResult(ctx, contextID, threadID, taskID, result.Success, output, result.PRUrl)
 }
--- a/internal/executor/runner.go
+++ b/internal/executor/runner.go
@@ -370,16 +370,19 @@ type ToolResultContent struct {
 
 // progressState tracks execution phase for compact progress reporting
 type progressState struct {
-	phase        string   // Current phase: Exploring, Implementing, Testing, Committing
-	filesRead    int      // Count of files read
-	filesWrite   int      // Count of files written
-	commands     int      // Count of bash commands
-	hasNavigator bool     // Project has Navigator
-	navPhase     string   // Navigator phase: INIT, RESEARCH, IMPL, VERIFY, COMPLETE
-	navIteration int      // Navigator loop iteration
-	navProgress  int      // Navigator-reported progress
-	exitSignal   bool     // Navigator EXIT_SIGNAL detected
-	commitSHAs   []string // Extracted commit SHAs from git output
+	phase        string // Current phase: Exploring, Implementing, Testing, Committing
+	filesRead    int    // Count of files read
+	filesWrite   int    // Count of files written
+	commands     int    // Count of bash commands
+	hasNavigator bool   // Project has Navigator
+	navPhase     string // Navigator phase: INIT, RESEARCH, IMPL, VERIFY, COMPLETE
+	navIteration int    // Navigator loop iteration
+	navProgress  int    // Navigator-reported progress
+	exitSignal   bool   // Navigator EXIT_SIGNAL detected
+	// A success-claiming exit signal with zero commits classifies as a decline, never done.
+	exitSignalSuccess bool
+	exitSignalReason  string
+	commitSHAs        []string // Extracted commit SHAs from git output
 	// Metrics tracking (TASK-13)
 	tokensInput              int64  // Input tokens used
 	tokensOutput             int64  // Output tokens used
@@ -1599,6 +1602,34 @@ func truncateDiagnostic(s string, max int) string {
 // parseDeclinedReason extracts the reason from a DECLINED:<reason> marker
 // emitted by Claude when a task is explicitly unactionable. Returns the reason
 // and true if found, or ("", false) if no marker is present. GH-2777.
+// declineReasonFromRun reports an explicit first-pass decline: a DECLINED: marker or a success-claiming exit signal.
+func declineReasonFromRun(backendResult *BackendResult, state *progressState) (string, bool) {
+	if backendResult != nil {
+		if reason, ok := parseDeclinedReason(strings.TrimSpace(backendResult.LastAssistantText)); ok {
+			return reason, true
+		}
+	}
+	if state != nil && state.exitSignal && state.exitSignalSuccess {
+		reason := state.exitSignalReason
+		if reason == "" {
+			reason = "executor signalled successful completion with no commit — nothing to change"
+		}
+		return reason, true
+	}
+	return "", false
+}
+
+func markDeclined(result *ExecutionResult, backendResult *BackendResult, reason string) {
+	result.Success = false
+	result.Declined = true
+	result.Error = ""
+	result.DeclinedReason = reason
+	result.Outcome = "declined" // TASK-358
+	if backendResult != nil {
+		backendResult.ErrorType = string(ErrorTypeDeclined)
+	}
+}
+
 func parseDeclinedReason(text string) (string, bool) {
 	const marker = "DECLINED:"
 	idx := strings.Index(text, marker)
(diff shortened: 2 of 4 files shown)


lkshrk:
Pushed an update after live testing surfaced a second first-pass decline route the marker check cannot see.

Observed on a watch-only ticket: the model produced a correct no-op rationale and a structured exit signal —

```
{"v":2,"type":"exit","exit_signal":true,"success":true}
```

— but no literal `DECLINED:` line, because **the first execution is never taught the DECLINED convention** (only the retry prompt introduces it, GH-2777). The ghost-SHA guard (GH-3126) then rightly refused to record the base-branch SHA as shipped, the run fell through to the no-commit retry, and the user got ❌ again — the exact symptom this PR removes, via a different route.

New behavior: `exit_signal` with `success:true` plus zero commits classifies as a **decline** — never "done" (the ghost guard's anti-ghost-close purpose stands), never "failed". The signal's reason/message becomes the decline reason. Second guard test added driving this route through `Execute`.

Also updated the issue-facing note: gh pr view for the diff; commit message carries the full three-gap narrative.

lkshrk:
Second live-test round showed the previous revision still failed on the real path — updated again, and this time the fix moved to where the failure actually happens.

The observed trace:
```
23:09:51 WARN  executor: harvested SHA is already on base branch — no new commit
23:09:51 ERROR Task execution failed  error="no new commit produced — worktree HEAD matches base branch parent"
```
No decline check ever ran: the ghost-SHA guard flips the run into the **failed branch**, where ERROR + a `task_failed` alert fire unconditionally — the decline checks lived in the success branch and were structurally unreachable for this shape. The false alert is arguably worse than the ❌: it trains operators to ignore the channel that exists for real breakage.

Two additional changes:

1. The guard's no-op failure is re-examined against the decline signal before the failed branch runs; a declined result takes a distinct terminal path (no ERROR, no alert, recorder finishes "declined"). The ghost guard's anti-ghost-close purpose is untouched — declined is never "done".
2. `ParseSignals` only read fenced ```pilot-signal``` blocks, but the model emits the payloads as **bare** `{"v":...}` lines — so the exit-signal capture added in the previous revision never fired in production. Bare lines now parse too (fenced blocks removed first to avoid double-count).

New guard test drives the exact observed shape: harvested base-branch SHA + bare exit signal → declined, no error, one backend call.

lkshrk:
Live confirmation from the deployment that hit the defect: on a build carrying this revision, the same watch-only ticket flow that previously rendered `❌ Task failed` plus a `task_failed` alert (twice, across two earlier revisions — see the thread above) now produces `⏸ No changes needed` with the model's stated reason, a `Task declined by executor` log line, and no alert. All three decline routes (DECLINED: marker, fenced exit signal, bare exit signal via the ghost-SHA path) are covered by tests, and the observed bare-signal shape is verified end to end.

alekspetrov (changes requested):
Splitting this review: two of the three legs are correct and welcome — please keep both. The first-pass `DECLINED:` marker check is sound (explicit opt-in marker, mandatory non-empty reason), and the comms rendering fix is real (main erases `Declined` and renders ❌ Task failed).

Requesting changes on the exit-signal leg. `{"v":2,"type":"exit","exit_signal":true,"success":true}` is the *mandatory* completion signal (workflow.go Phase 5, "REQUIRED") — it's emitted on every run the model believes succeeded, including the GH-916 "claimed success, never committed" class (~10% of failures) this pipeline explicitly defends against. Treating it as a decline:

- renders forgot-to-commit failures as "⏸ No changes needed",
- makes the no-commit retry effectively unreachable in production (existing retry tests pass only because their fake backends emit no signals), and
- runs before `preserveDirtyWorktreeAsWIP`, so real uncommitted work gets deleted and reported as an intentional no-op — the exact incident GH-4517 fixed.

Minimum bar for classifying an exit-signal no-commit as a decline: the worktree must be verified clean (run the GH-4517 preserve check first), and the signal must carry an explicit reason — no fabricated default. Better: a dedicated opt-in marker (e.g. `"no_op": true` + mandatory reason) instead of overloading the routine completion signal. We're speccing that contract on our side and would gladly review a PR against it.

Separately: please drop or split out the bare-line JSON parsing in signal.go — it widens the signal surface for all types with no signal_test.go coverage, and it's what makes the routine signal trip this path. Smaller items: the comms decline branch uses `SendText`, dropping `threadID` (use the thread-correct `SendChunked`); `markDeclined`/`finishDeclined` duplicate the same mutations; ghost-path declines record `failed` in metrics while retry-path declines record `success`.

alekspetrov:
Heads-up: the safe no-op contract from the review is now queued as #4964. It salvages this PR's legs 1+2 (shared finishDeclined dedup + first-pass DECLINED marker; ⏸ rendering) re-sequenced behind preserve-before-classify, and replaces the exit-signal inference (leg 3) with an explicit `no_op:true`+reason signal. If you'd rather carry it yourself against that spec, comment there — otherwise the queue will take it.

alekspetrov:
Status: #4875 is now closed — the decline contract landed via #4977 (GH-4964) and has been in production since v2.262.0: decline honored before the no-commit retry, ⏸ no-op rendering, preserve-before-classify. That covers this PR's legs 1+2 as re-sequenced.

This PR is currently CONFLICTING against that landing. If your live deployment still shows residuals #4977 missed — your bare-unfenced-signal parser point is the likely candidate — a rebase down to just that residual would be very welcome. Otherwise feel free to close; either way, your live-fire confirmations on this thread were what pinned the contract's spec. Thanks.

Issue #960, opened 2026-02-13
Wire signal parser v2 for progress tracking


alekspetrov:
Parent: GH-951

Integrate signal.go parser into runner.go execution path (GH-941)
Signal parser v2 exists but isn't wired into the main execution loop. This enables proper progress extraction during worktree execution, important for autopilot tracking and user feedback.

alekspetrov:
✅ Completed as part of GH-951
PR: https://github.com/alekspetrov/pilot/pull/965

PR #4903, opened 2026-08-15
fix(comms): strip pilot-signal protocol blocks from user-facing replies

### LINKED ISSUE #4902: pilot-signal protocol blocks leak into user-facing chat replies
### Summary

The model's `pilot-signal` protocol output — fenced ```pilot-signal``` blocks and bare `{"v":2,...}` JSON lines — reaches the user verbatim in chat replies. Observed on Telegram: a task reply contained `{"v":2,"type":"status","phase":"INIT","progress":5}` and the exit signal `{"v":2,"type":"exit","exit_signal":true,"success":true}` inline in the message.

`CleanInternalSignals` (`internal/comms/util.go`) strips only the legacy string markers (`EXIT_SIGNAL: true`, `[NAV_COMPLETE]`, ...) and predates the structured v2 signal protocol (GH-960). The signal *parser* consumes the blocks for progress tracking, but nothing removes them from the text handed to messengers.

Related: there are four divergent `CleanInternalSignals` copies (comms, telegram, slack, discord formatters) — each strips a different signal set; none knows the v2 protocol.

### Steps to Reproduce

1. Any chat-originated task whose model output includes pilot-signal blocks (standard for signal-emitting prompts).
2. Read the reply in Telegram.

### Expected Behavior

Protocol blocks are internal plumbing; the user sees only prose.

### Actual Behavior

Raw protocol JSON rendered inside the chat message.

### Pilot Version

v2.260.6 (fork build of main @ 735c938f; affected paths unmodified from upstream)

### Adapter

Telegram (observed; the comms cleaner is adapter-independent)

lkshrk:
`CleanInternalSignals` predates the v2 signal protocol (GH-960): it strips legacy string markers but passes fenced ```pilot-signal``` blocks and bare `{"v":...}` JSON lines straight through — chat replies arrived carrying raw protocol payloads inline (observed live on Telegram, see issue).

Comms cleaner + the telegram formatter copy now drop pilot-signal fences and bare protocol lines; ordinary code fences and JSON in prose are untouched (tests cover both keep-cases). Slack/discord formatter copies share the gap — noted in the issue; the four-way cleaner divergence is worth its own consolidation pass.

Fixes #4902

Verified: comms + telegram suites green, `golangci-lint --new-from-rev=main` 0 issues.

### DIFF
--- a/internal/adapters/telegram/formatter.go
+++ b/internal/adapters/telegram/formatter.go
@@ -336,8 +336,23 @@ func cleanInternalSignals(text string) string {
 	lines := strings.Split(text, "\n")
 	var cleanLines []string
 	skipBlock := false
+	inSignalFence := false
 
 	for _, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if inSignalFence {
+			if trimmed == "```" {
+				inSignalFence = false
+			}
+			continue
+		}
+		if strings.HasPrefix(trimmed, "```pilot-signal") {
+			inSignalFence = true
+			continue
+		}
+		if strings.HasPrefix(trimmed, `{"v":`) && strings.HasSuffix(trimmed, "}") {
+			continue
+		}
 		// Skip NAVIGATOR_STATUS blocks
 		if strings.Contains(line, "NAVIGATOR_STATUS") {
 			skipBlock = true
--- a/internal/comms/util.go
+++ b/internal/comms/util.go
@@ -29,8 +29,23 @@ func CleanInternalSignals(text string) string {
 	lines := strings.Split(text, "\n")
 	var clean []string
 	skipBlock := false
+	inSignalFence := false
 
 	for _, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if inSignalFence {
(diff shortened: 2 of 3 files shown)


alekspetrov (approved):
LGTM — the stripping logic matches `signalBlockRegex` exactly (TrimSpace makes it strictly more lenient than the producer, which is the safe direction), and the keep-cases are covered. Two asks before merge:

1. Add the same fence/bare-line cases to the telegram formatter's test — only `comms/util_test.go` grew cases, and untested parallel copies are exactly how these implementations drift.
2. A comment on the fence loop noting that an unterminated ```pilot-signal block intentionally strips to EOF — right trade-off, but it should be visibly deliberate.

On the slack/discord follow-up you scoped out: we dug into it, and it's worse than drift — both adapters' `FormatTaskResult` never call their `CleanInternalSignals` at all (dead code, zero production call sites), so signals leak there unconditionally today. We're planning a consolidation onto `comms.CleanInternalSignals` on our side, landing after this PR so all callers inherit the fence stripping.

Issue #4244, opened 2026-07-12
refactor(executor,autopilot): unify recordExecutionEvent wrappers on validate-first — make FK-787 unrepresentable


alekspetrov:
## Context

Four independent `recordExecutionEvent` wrappers exist, each duplicating nil-store/skip guards, and only one validates that the parent `executions` row exists before inserting: `Runner.recordExecutionEvent` (`internal/executor/runner.go:1294`, ~24 call sites — uses `task.LogExecutionID()` whose bare-`task.ID` fallback is the FK-787 generator), `Dispatcher.recordExecutionEvent` (`internal/executor/dispatcher.go:459`), `ProjectWorker.recordExecutionEvent` (`internal/executor/dispatcher.go:1051`), and `Controller.recordExecutionEvent` (`internal/autopilot/controller.go:719` — the only validate-first implementation: `GetLatestExecutionByTaskID`, warn-and-skip when missing).

With #4243's `ExecutionLifecycle` in place, every legitimate execution has a row — so a missing row at event-write time is always a bug that should log loudly, never an FK constraint error.

Blocked by: #4243

## Approach

Single-scope change (do not decompose):

1. Promote the Controller's validate-first pattern into one shared implementation (natural home: a method on `ExecutionLifecycle` from #4243, or `internal/memory` next to `InsertExecutionEvent` at `store.go:779`).
2. Behavior: resolve the execution ID (verify the row exists); if missing → `slog.Warn` with task_id/stage/detail and skip the insert (fail-loud, never FK-fail); if present → insert.
3. Migrate all four wrappers to delegate to it; delete the duplicated guard bodies.
4. Keep `task.LogExecutionID()` fallback semantics for resolution order (ExecutionID → ParentExecutionID → task.ID) but the task.ID fallback now goes through the existence check instead of a blind insert.

## Acceptance Criteria

- [ ] One shared event-write implementation; the four wrappers are thin delegates (or callers migrated outright).
- [ ] Writing an event for a nonexistent execution produces a Warn log and no error — regression test proves FK-787 is impossible through any wrapper.
- [ ] Existing event-write behavior for valid rows unchanged (existing `execution_events` tests green: `internal/memory/execution_events_test.go`, `internal/autopilot/controller_execution_events_test.go`).
- [ ] `go test -race ./...` green.

## Refs

- Blocked by: #4243 (ExecutionLifecycle chokepoint)
- Validate-first template: `internal/autopilot/controller.go:719`
[...]

Issue #4243, opened 2026-07-12
refactor(executor,memory): ExecutionLifecycle chokepoint — single Begin/Transition/Finish API for execution rows


alekspetrov:
## Context

Execution-row lifecycle is hand-rolled at every call site: three production paths construct `memory.Execution` structs and call `SaveExecution` directly (`internal/executor/dispatcher.go:547,605`, `internal/executor/epic.go:1988`), transitions are scattered across `MarkExecutionCompleted` ×4 / `UpdateExecutionStatus` ×6 sites, and the status vocabulary exists only as free-text strings plus an `if` chain (`internal/memory/store.go:1371`). Every time a path forgets a step we get a production defect class: FK-787 on `execution_events` (TASK-394 epic path; #4205 CLI path), invisible metrics, false dashboard status (TASK-399). The fix pattern so far has been guards at call sites (#4229 extended one guard to 4 places) — this issue moves the invariant into the type system instead.

Blocked by: #4205

## Approach

Single-scope refactor (do not decompose):

1. New `ExecutionLifecycle` type (suggest `internal/executor/lifecycle.go`), thin over `*memory.Store`:
   - `Begin(task *Task, initial Status) (execID string, err error)` — creates the row AND threads `execID` into `task.ExecutionID` (the step `cmd/pilot/commands.go` historically missed; `buildTaskFromExecution` at `dispatcher.go:1105` shows the correct threading).
   - `Transition(execID string, s Status) error` — non-terminal moves (queued→running).
   - `Finish(execID string, result *ExecutionResult) error` — wraps `TerminalStatus` (`runner.go:232-256`) + `MarkExecutionCompleted`/`UpdateExecutionStatus` + `SaveExecutionMetrics`, generalizing `finalizeSubIssueExecution` (`runner.go:1809-1845`).
2. Introduce a Go const set for the status vocabulary (`StatusQueued`, `StatusRunning`, `StatusCompleted`, `StatusFailed`, `StatusNoOp`, `StatusStalled`, `StatusRateLimited`, `StatusInfra`, `StatusSkipped`, `StatusDeclined`, `StatusDecomposed`) — same strings, typed. No DB schema change.
[...]

PR #4904, opened 2026-08-15
fix(comms): give chat-originated tasks an executions row via ExecutionLifecycle

### LINKED ISSUE #4878: Execution events silently dropped with 'execution not found' — including gh-guard denials, leaving no audit trail
### Summary

Every execution event for a task can be silently dropped with `execution not found: execution_id=<id>` — the whole lifecycle audit trail (`spec_validated`, `claude_started`, `implementation_started`) **and security-relevant events** (`executor.gh_guard_denied`) alike. Each drop logs a Warn and execution continues, so the task runs and completes normally while the events table records nothing about it.

The guard case is the sharpest consequence: a security guard whose denials are unrecorded cannot be reviewed after the fact. But the failure is upstream of the guard — the events writer references an execution row that does not exist under that id, so *nothing* about the task is auditable.

### Steps to Reproduce

Observed on a Telegram-originated task (`TG-1786817399`). Every `Failed to record execution event` in the trace carries the same `execution not found` error, suggesting the execution row was never created under the id the event writer uses (id-scheme mismatch between the comms/Telegram task path and the executions ledger, rather than a race on one event).

### Expected Behavior

Either the execution row exists before the first event is recorded, or event recording fails loudly enough to be noticed — at minimum, guard-denial events should never be droppable on a missing-row condition (create the row on first write, or buffer and flush).

### Actual Behavior

All four lifecycle events for the task dropped with identical errors; task proceeded normally; no audit trail exists.

### Pilot Version

v2.260.3 (fork build of main @ 735c938f; executor/ledger paths unmodified from upstream)

### Adapter

Telegram (task origin; the recording path is adapter-independent)

### Logs

```
stage=spec_validated            error="execution not found: execution_id=TG-1786817399"
stage=claude_started            error="execution not found: execution_id=TG-1786817399"
stage=implementation_started    error="execution not found: execution_id=TG-1786817399"
stage=executor.gh_guard_denied  error="execution not found: execution_id=TG-1786817399"
```

lkshrk:
Chat tasks (`executeTaskCore`) call `runner.Execute` directly and never touch the dispatcher or the `ExecutionLifecycle` chokepoint, so no `executions` row exists under the id runner-side writes key on. Every `execution_events` write for a TG-*/SLACK-* task — `spec_validated`, `claude_started`, `implementation_started`, and security-relevant `executor.gh_guard_denied` — was dropped by the validate-first guard (GH-4244) with `execution not found`. A gh-guard denial therefore left **no audit trail at all** (observed live, see #4878).

`executeTaskCore` now mirrors the CLI path (GH-4243's `recordCLITaskStart`/`recordCLITaskFinish`): `Begin` stamps `task.ExecutionID` and creates the claimed row before `Execute`; `Finish` records terminal status and metrics after. A lost claim aborts before backend invocation with a visible message; any other lifecycle error degrades to today's behavior (execute without audit trail) rather than blocking the task.

Question/research/chat-conversation paths stay row-less — ephemeral by design, no lifecycle stages emitted; scoped to the task path where the drops were observed.

Fixes #4878

Verified: comms + executor suites green, `golangci-lint --new-from-rev=main` 0 issues. Guard test drives `executeTask` with a real store and asserts one finalized row and successful event writes against it.

### DIFF
--- a/internal/comms/handler.go
+++ b/internal/comms/handler.go
@@ -3,6 +3,7 @@ package comms
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strings"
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
Approve — this is exactly the Begin/Finish shape the lifecycle chokepoint was built for. Initial status, (task, project, generation) scoping, and the ErrClaimLost abort-before-backend contract all match the CLI precedent, and Finish correctly gates on a row actually existing. The end-to-end test asserts the real defect (event writes landing on a finalized row), not a mock of it.

If you get a chance: a small test for the ErrClaimLost branch (pre-claim the (taskID, projectPath, 0) key and assert the abort) would complete the coverage. The dangling "Starting…" progress message next to the abort text is cosmetic.

PR #4905, opened 2026-08-15
fix(ghguard): honor an explicit -X GET as a read even with data flags

### LINKED ISSUE #4877: gh-guard denies read-only 'gh api search/* -f q=' as a mutation
### Summary

gh-guard denies documented **read-only** GitHub API forms. `gh api search/issues -f q=...` is the CLI's documented way to pass a search query, and search endpoints are GET-only — but `checkAPIRead` (`internal/executor/ghguard/policy.go:396`) treats the mere presence of `-f`/`-F`/`--field`/`--raw-field`/`--input` as a mutation and denies.

The code comment says this is deliberate ("stricter than the letter of gh api non-GET... closes that implicit-mutation gap deliberately") — this issue is about the cost of that tradeoff: executors doing legitimate research (checking whether an upstream issue has a linked PR, here `gh api search/issues -f q='6220 in:body type:pr'`) are blocked from the only ergonomic form of a read-only query, and the denial reason ("carries a request body and is treated as a mutation") tells the model its read was a write.

### Steps to Reproduce

1. In an executor session, run `gh api search/issues -f q='anything'`.
2. Guard denies with the request-body reason.

### Expected Behavior

Read-only endpoints stay usable. Options, least to most invasive:

- Allowlist known GET-only endpoint families (`search/*`) even with data flags — gh sends `q` as a query parameter for GETs, not a body.
- Honor an explicit `-X GET` as an override: data flags plus an explicit GET method is unambiguous (gh only auto-switches to POST when `-X` is absent).
- At minimum, make the denial message suggest the guard-safe alternative (`gh api 'search/issues?q=...'` URL-encoded) so the model can self-correct.

### Actual Behavior

Denied as a mutation; executor loses the search capability entirely unless it discovers URL-encoding on its own.

### Pilot Version

v2.260.3 (fork build of main @ 735c938f; ghguard unmodified from upstream)

### Adapter

GitHub

### Logs

```
18:13:12 WARN  gh_guard_denied component=executor.ghguard_audit task_id=TG-1786817399 repo=""
               args="api search/issues -f q=... | {number, title, state, updatedAt}"
               reason="gh api with -f/-F/--field/--raw-field/--input carries a request body and is treated as a mutation"
```

lkshrk:
The data-flag heuristic denies any `gh api` call carrying `-f/-F/--field/--raw-field/--input` to close the implicit-mutation gap (gh auto-switches to POST when data flags are present). But gh only auto-switches **when `-X` is absent** — data flags plus an explicit `-X GET` are unambiguously a query-parameter read. The blanket deny blocked the CLI's documented search form (`gh api search/issues -f q=...`), cutting executors off from read-only research (observed live, #4877).

Change: data flags + explicit GET now pass; the deny message for the remaining ambiguous case teaches both escapes (add `-X GET`, or URL-encode as `gh api 'search/issues?q=...'`) so the model can self-correct. Implicit-mutation behavior without `-X` is unchanged — the deliberate strictness the `checkAPIRead` comment documents still holds where it is actually ambiguous.

Fixes #4877

Verified: ghguard + executor suites green, `golangci-lint --new-from-rev=main` 0 issues. Test rows added for both explicit-GET spellings.

### DIFF
--- a/internal/executor/ghguard/policy.go
+++ b/internal/executor/ghguard/policy.go
@@ -392,10 +392,14 @@ func Classify(id Identity, args []string) Decision {
 // presence as non-GET is stricter than the letter of "gh api non-GET" but
 // closes that implicit-mutation gap deliberately.
 func checkAPIRead(p parsedArgs) Decision {
-	if p.hasDataFlag {
-		return deny("gh api with -f/-F/--field/--raw-field/--input carries a request body and is treated as a mutation", allowedSummary())
-	}
 	method := strings.ToUpper(strings.TrimSpace(p.method))
+	if p.hasDataFlag {
+		// gh auto-switches to POST only when -X is absent; explicit GET + data flags is a query-parameter read.
+		if method == "GET" {
+			return Decision{Verdict: VerdictAllow, Reason: "gh api explicit -X GET with query fields"}
+		}
+		return deny("gh api with -f/-F/--field/--raw-field/--input carries a request body and is treated as a mutation; for a read-only query add an explicit -X GET, or URL-encode it (gh api 'search/issues?q=...')", allowedSummary())
+	}
 	if method != "" && method != "GET" {
 		return deny(fmt.Sprintf("gh api -X %s is not a read", p.method), allowedSummary())
 	}
--- a/internal/executor/ghguard/policy_test.go
+++ b/internal/executor/ghguard/policy_test.go
@@ -43,6 +43,8 @@ func TestClassify(t *testing.T) {
 		{"api GET implicit", []string{"api", "repos/qf-studio/pilot/issues/1"}, VerdictAllow},
 		{"api GET explicit -X", []string{"api", "repos/qf-studio/pilot/issues/1", "-X", "GET"}, VerdictAllow},
 		{"api GET explicit --method lowercase", []string{"api", "repos/qf-studio/pilot/issues/1", "--method", "get"}, VerdictAllow},
+		{"api search with -f query and explicit GET", []string{"api", "search/issues", "-X", "GET", "-f", "q=6220 in:body type:pr"}, VerdictAllow},
+		{"api search with -f query lowercase get", []string{"api", "search/issues", "--method", "get", "-f", "q=repo:o/r is:open"}, VerdictAllow},
 
 		// --- own-artifact allows ---
 		{"pr create no head (current branch)", []string{"pr", "create", "--title", "t", "--body", "b"}, VerdictAllow},


alekspetrov (changes requested):
The premise checks out — gh auto-switches to POST only when `-X` is absent, explicit GET moves fields to the query string, and the #4877 cost argument is real. But the relaxation is only as safe as the guard's method extraction being faithful to gh's, and it isn't: gh uses pflag, which accepts concatenated shorthand (`-XPOST` ≡ `-X POST`) with last-one-wins, and `parseArgs` silently ignores those tokens. Concretely:

`gh api repos/o/r/issues/123 -X GET -XPATCH -f state=closed`

classifies as allow under this PR ("explicit -X GET with query fields") while gh executes a PATCH with a body — verified against both Classify and live api.github.com. That's the exact hostile-prompt smuggling this guard exists to stop.

Asks:
1. Handle pflag's concatenated shorthands in parseArgs (`-XPOST`, `-fk=v`, `-X=GET`) with last-occurrence-wins matching gh.
2. Keep `--input` denied even with `-X GET` — per gh's docs it is always the request body (only fields move to the query string), so it contradicts the "unambiguously a query-parameter read" rationale.
3. Add deny regression rows for the override and concatenated forms (`-X GET -XPATCH -f …`, `-XPUT`, `-fq=…`).

Full disclosure: while verifying this we found the same parse gap already allows body-less mutations on main today (`-XPUT` alone classifies as a read), plus several related shapes — so we're speccing a broader parser-parity hardening of ghguard on our side. Your GET+fields semantics are correct and we want them (GraphQL-over-GET verified inert, though we'll belt-and-braces that too); once the parser is pflag-faithful this relaxation can land on top, either amended here or rebased onto the hardened parser.

alekspetrov:
Heads-up: the hardened parser this review asked for is now queued as #4963 (pflag-faithful per-command specs), and the env-bypass sibling as #4968. Your explicit-GET relaxation is wanted on top of #4963 — its spec includes your #4877 test rows plus the `--input`-is-always-a-body carve-out. Feel free to rebase onto it once it lands, or we'll fold the relaxation in per the spec if this PR goes quiet.

alekspetrov:
Status: the hardened parser landed — #4963 → PR#4975 (pflag-faithful per-command specs) and the env-bypass sibling #4968 → PR#4981 both merged 08-18, in production since v2.262.0. Good news: the explicit-GET relaxation this PR proposed went in **with** #4975 — `checkAPIRead` honors `-X GET`/`--method GET` with data flags as a read, and the code comment credits #4877/#4905. So the behavior you wanted is live and this PR is now CONFLICTING against it. Closing the loop: nothing left to rebase unless you see a case the new parser still misclassifies — if so, please file it against the #4975 parser and we'll take it. Thanks for pushing on this; the relaxation shipped because of these reports.

PR #4907, opened 2026-08-15
fix(telegram): surface edited-message and empty-text drops instead of hiding them

### LINKED ISSUE #4906: Telegram: edited messages and empty-text updates die silently (edited_message not parsed)
### Summary

`Update` (`internal/adapters/telegram/client.go`) parses only `message` and `callback_query`. An edited message arrives as `edited_message`, so `processUpdate` hits `update.Message == nil` and returns silently — no log line at any level. A user who edits their message to fix a typo gets nothing, and the operator has no trace the update ever arrived. The empty-`Text` drop a few lines later is equally silent.

Diagnosing this class costs a daemon-stop + raw `getUpdates` listener session (how it was found).

### Steps to Reproduce

1. Send the bot any message; edit it.
2. Nothing happens; nothing is logged.

### Expected Behavior

At minimum, both silent returns log with the update's identifying fields so a dropped update is diagnosable from the daemon log. Whether an edit should be *re-processed* as a fresh message is a separate product decision (double-dispatch risk for task-triggering edits) — this issue only asks for the silence to end; a fix can parse `edited_message` and explicitly log that edits are ignored.

### Actual Behavior

Silent return; the update evaporates.

### Pilot Version

v2.260.7 (fork build of main @ 735c938f; telegram update parsing unmodified from upstream)

### Adapter

Telegram

lkshrk:
`Update` parsed only `message` and `callback_query`; an `edited_message` hit the `Message == nil` return and evaporated with no log line at any level (found via a daemon-stop + raw `getUpdates` session — see #4906). The empty-`Text` drop was equally silent.

`Update` gains `EditedMessage`. Edits log at Info and are deliberately **not** re-processed — a task-triggering edit would double-dispatch; re-routing edits as fresh messages is a separate product decision the issue leaves open. The no-payload and empty-text drops log at Debug with identifying fields.

Fixes #4906

Verified: telegram suite green, `golangci-lint --new-from-rev=main` 0 issues. Tests cover the parse (`edited_message` → `Update.EditedMessage`, `Message` stays nil) and the no-dispatch path.

### DIFF
--- a/internal/adapters/telegram/client.go
+++ b/internal/adapters/telegram/client.go
@@ -69,6 +69,7 @@ type Result struct {
 type Update struct {
 	UpdateID      int64          `json:"update_id"`
 	Message       *Message       `json:"message,omitempty"`
+	EditedMessage *Message       `json:"edited_message,omitempty"`
 	CallbackQuery *CallbackQuery `json:"callback_query,omitempty"`
 }
 
--- a/internal/adapters/telegram/handler.go
+++ b/internal/adapters/telegram/handler.go
@@ -310,6 +310,20 @@ func (h *Handler) processUpdate(ctx context.Context, update *Update) {
 	}
 
 	if update.Message == nil {
+		// Edits are deliberately not re-processed: a task-triggering edit would double-dispatch.
+		if update.EditedMessage != nil {
+			chatID := int64(0)
+			if update.EditedMessage.Chat != nil {
+				chatID = update.EditedMessage.Chat.ID
+			}
+			logging.WithComponent("telegram").Info("ignoring edited message",
+				slog.Int64("chat_id", chatID),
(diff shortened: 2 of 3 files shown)


alekspetrov (approved):
Confirmed — `edited_message` deserializes with `Message == nil` and evaporates with zero log output at any level. The minimal fix (parse the field, log, explicitly don't dispatch) is the right call, and stating the double-dispatch risk instead of sneaking re-processing in is appreciated.

One nit: both new log lines fire before the `allowed_ids` check, so any stranger DMing the bot gets their edits logged at Info, one line per edit, no dedupe. Main already runs photo/voice handling before the allowlist check, so this follows existing ordering — but consider Debug for edits from non-allowed chats, or a comment accepting the noise. Non-blocking; approving.

Issue #4964, opened 2026-08-18
fix(executor): dedicated no_op decline signal — never infer a decline from the mandatory exit signal


alekspetrov:
no-decompose

Single coherent change (workflow.go signal contract + signal.go parser + runner.go ordering + comms) — do NOT decompose; one PR.

## Problem

There is no way for the executor to say "I looked, and correctly nothing needed to change." PR #4901 tried to infer it from `state.exitSignalSuccess` + zero commits — but `{"v":2,"type":"exit","exit_signal":true,"success":true}` is the **mandatory** completion signal (`workflow.go:169-172`, Phase 5 "REQUIRED"), emitted on every run the model *believes* finished. That includes the GH-916 class (model claims success, never committed — ~10% of failures per `runner.go:3967`). Inferring a decline from it would:

1. render forgot-to-commit failures as "⏸ No changes needed",
2. make the GH-916 retry unreachable in production, and
3. run before `preserveDirtyWorktreeAsWIP`, deleting real uncommitted work — the pilot-console#26/B8 incident GH-4517 fixed.

The signal vocabulary has a *failure* exit (`success:false, reason:"blocked: …"`, `workflow.go:210-213`) but nothing symmetric for "succeeded because nothing was needed."

Also found: **`DECLINED:` is prompted nowhere in the first-pass prompt.** It appears only in the GH-916 *retry* prompt (`runner.go:4008-4021`, "Option B — Decline"). `workflow.go`'s Phase 5 contract never mentions it, so first-pass decline handling only fires when the model emits the marker spontaneously.

## Current flow (main, `runner.go` `Execute()`)

| Step | Lines | Behavior |
|---|---|---|
| CommitSHA harvest fallbacks | 3805-3844 | git-log → `GetCurrentCommitSHA` → structured summary, only if `CommitSHA == ""` |
| Ghost-SHA guard | 3850 → `git_freshness.go:23-68` | **Internally correct**: `preserveDirtyWorktreeAsWIP` runs first (`:54`); only if nothing preserved does it set `Error = "no new commit produced…"` (`:63-65`). Skipped when `CommitSHA == ""` |
| Metrics | 3882-3897 | ⚠ `RecordExecution` label derives from `result.Success` **before** decline/no_op classification — pre-existing gap, declines count as `success` |
| Success/failure branch | 3900 / 3939 | |
[...]

PR #4977, opened 2026-08-18
fix(executor): dedicated no_op decline signal — never infer decline from mandatory exit signal

alekspetrov:
## Summary

- The mandatory exit signal (`{"type":"exit","exit_signal":true,"success":true}`) fires on every run the model believes finished, including the GH-916 failure class where it claims success but never commits (~10% of failures). A prior external attempt (#4901, never merged) inferred a no-op decline from this bare signal, which would have mislabeled forgot-to-commit failures as "no changes needed" and made the GH-916 retry unreachable.
- Adds an opt-in `no_op:true` + mandatory `reason` field to the exit signal contract (`workflow.go`, `signal.go`) as the **only** new evidence for a decline, alongside the existing `DECLINED:<reason>` text marker.
- Routes every decline insertion point in `runner.go`'s `Execute()` (ghost-SHA rejection, GH-916 pre-retry, GH-916 post-retry) through shared `finishDeclined`/`preserveDirtyOrFail` helpers, guaranteeing the GH-4517 dirty-worktree auto-preserve backstop always runs and wins before any decline is honored — a real uncommitted diff always outranks a no-op claim, closing a latent post-retry ordering gap.
- Comms now renders a decline via `SendChunked` (threaded, "⏸ No changes needed: …"), not as a failure via `SendResult`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite)
- [x] `scripts/check-secret-patterns.sh` / `scripts/check-mocks.sh` — clean
- [x] New unit tests: `signal_test.go` (`NoOpExitReason` — 8 cases), `handler_test.go` (declined-vs-normal comms rendering — 3 cases)
- [x] New integration tests via `Runner.Execute()` in `gh4964_test.go`: bare-signal retry (regression guard), no_op+reason clean-tree decline, no_op+reason dirty-tree preserve-wins, ghost-SHA+no_op decline, ghost-SHA+dirty preserve-wins, post-retry DECLINED+dirty preserve-wins

### DIFF
--- a/internal/comms/handler.go
+++ b/internal/comms/handler.go
@@ -869,6 +869,25 @@ func (h *Handler) executeTaskCore(ctx context.Context, contextID, threadID, task
 	}
 
 	output := CleanInternalSignals(result.Output)
+	h.sendTaskResult(ctx, contextID, threadID, taskID, result, output)
+}
+
+// sendTaskResult renders a finished ExecutionResult to chat. A declined task
+// (GH-4964: DECLINED marker or explicit no_op+reason exit signal — never the
+// bare mandatory exit signal) is not a failure and must not render as ❌ via
+// SendResult's success=false path; it gets its own "no changes needed"
+// message via SendChunked, which — unlike SendText — carries threadID so
+// Slack replies land in the originating thread (ThreadTS) instead of the
+// channel root; Telegram/Discord ignore threadID harmlessly.
+func (h *Handler) sendTaskResult(ctx context.Context, contextID, threadID, taskID string, result *executor.ExecutionResult, output string) {
+	if result.Declined {
+		reason := result.DeclinedReason
+		if reason == "" {
+			reason = "no reason provided"
+		}
+		_ = h.messenger.SendChunked(ctx, contextID, threadID, fmt.Sprintf("⏸ No changes needed: %s", reason), "")
+		return
+	}
 	_ = h.messenger.SendResult(ctx, contextID, threadID, taskID, result.Success, output, result.PRUrl)
 }
 
--- a/internal/comms/handler_test.go
+++ b/internal/comms/handler_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 	"unicode"
 
+	"github.com/qf-studio/pilot/internal/executor"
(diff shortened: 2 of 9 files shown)


alekspetrov:
## Post-merge review — VERDICT: APPROVE-with-notes

Implementation matches the TASK-480/GH-4964 contract precisely — decline evidence is strictly `no_op:true`+non-empty reason or a DECLINED marker (never the bare mandatory exit signal), and `preserveDirtyWorktreeAsWIP` verifiably wins before classification at all three insertion points, including the fixed post-retry ordering bug. Tests exercise the real `Runner.Execute()` path against real git repos and assert the load-bearing behaviors (retry reachability, preserve-over-decline, pushed WIP commits, threaded comms rendering).

Notes (none blocking):
1. `runner.go:4227-4230` — the post-retry `no_op` check is unreachable dead code: the GH-916 retry's EventHandler never routes text through the signal parser, and a first-pass `no_op` with a clean tree already declines pre-retry. Harmless defensive symmetry.
2. `runner.go:6050-6055` — `exitSignalNoOp`/`exitSignalReason` are sticky across the smart-retry loop; a premature `no_op` followed by more work can decline with a stale reason — but only when clean-tree AND zero-commits, so exposure is a mislabeled reason string, not lost work.
3. `runner.go:1605-1623` — first-pass DECLINED detection is substring-based over `LastAssistantText`; prose quoting the marker could false-positive (gated behind clean-tree + zero-commits; sanctioned by the spec's #4901 salvage).
4. `gh4517_test.go:164-167` — deliberate behavior change: dirty tree pre-retry now auto-preserves as WIP instead of getting the GH-916 retry. Matches the spec's ordering diagram verbatim; trades possible clean self-recovery for a manual-review WIP.

Spec conformance verified in full: `NoOpExitReason` requires exit-type + `success` + `no_op` + non-empty trimmed reason with no fallback; ghost-SHA insertion gates on the exact `ghostSHACleanNoCommitError` const; issue decisions honored (metrics gap deferred, `no_op` name, no label reuse). All relevant tests green on main.

PR #4978, opened 2026-08-18
fix(linear): filter GetLabelByName by team UUID when team_id is a UUID

### LINKED ISSUE #4965: fix(linear): GetLabelByName must filter by team id when team_id is a UUID (unfixed read half of #4884)
## Context

Leg A of `.agent/tasks/TASK-481-review-followup-defects.md`. Gate cleared: #4896 is MERGED — implement the read-filter branch only, reusing its `looksLikeUUID`/`ResolveTeamUUID` helpers.

## Problem — Linear `GetLabelByName` fails for a UUID-configured `team_id`

**Defect.** `GetLabelByName` filters `team: { key: { eq: $teamId } }` unconditionally (`internal/adapters/linear/client.go:416`). Linear's `key` is the short slug (`"ROU"`), not the UUID. Config documents `team_id` as accepting either form (`internal/config/config.go:61,96` — "Team ID or name"). With a UUID configured, lookup never matches, `GetOrCreateLabel` falls through to `CreateLabel` on every poller startup, and Linear rejects the duplicate name.

**Reproduction path** (confirmed): `internal/adapters/linear/poller.go:187,195,200,205` — `cacheLabelIDs` passes `p.config.TeamID` straight into both `GetLabelByName` and `GetOrCreateLabel`. (Contrast `CreateIssue`, which resolves through `parent.Team.Key` — always a key, never hits this.)

**Relationship to PR #4896**: that PR fixes only the *write* half (`ResolveTeamUUID` before `issueLabelCreate.teamId`) and adds the `looksLikeUUID`/`ResolveTeamUUID` helpers this leg reuses, plus a workspace-label fallback at `client.go:435-442`. The read filter is untouched — this is the unfixed half of issue #4884.

**Fix.** Branch the GraphQL filter on `looksLikeUUID(teamID)`: UUID ⇒ `team: { id: { eq: $teamId } }`, else keep `team: { key: { eq: $teamId } }`. Two named queries sharing the response struct is cleaner than string-templating one query. Reuse #4896's helpers as-is. The `len(nodes)==0` → `getWorkspaceLabelByName` fallback stays as #4896 leaves it.

**Files**: `internal/adapters/linear/client.go` (`GetLabelByName`, 412-442) · `internal/adapters/linear/client_test.go`.

**Test rows** (httptest mock, following #4896's `TestGetLabelByNameWorkspaceFallback`):

| Case | teamID | Expected filter sent | Expected |
|---|---|---|---|
| Key-configured, label exists | `"ROU"` | `team.key.eq` | returns ID, no fallback |
| UUID-configured, label exists | UUID | `team.id.eq` | returns ID, no fallback |
| UUID-configured, only workspace-scoped label | UUID | id-filter then workspace fallback | returns workspace label ID |
| Key-configured, missing everywhere | `"ROU"` | key-filter then fallback | `label %q not found in team %s` |
| `GetOrCreateLabel` idempotency, UUID team | UUID, label exists | id-filter | **`CreateLabel` never called** (the #4884 duplicate-create guard) |

**Dependency**: rebase onto #4896 (adjacent lines in the same function). If #4896 hasn't merged, implement both in one PR but keep the diffs logically separable.

---

## Acceptance

- All five test rows above pass; with a UUID `team_id`, `GetOrCreateLabel` never calls `CreateLabel` when the label exists (the #4884 duplicate-create guard).

alekspetrov:
## Summary
- Read half of GH-4884, unfixed by #4896 (which only fixed the write path via `ResolveTeamUUID`).
- `GetLabelByName` filtered `team.key.eq` unconditionally, but Linear's `key` is the short slug ("ROU"), not a UUID. With a UUID-configured `team_id`, the lookup never matched, so `GetOrCreateLabel` fell through to `CreateLabel` on every poller startup and Linear rejected the duplicate name.
- Branches the GraphQL query on `looksLikeUUID(teamID)` (reused from #4896): UUID → `team: { id: { eq: $teamId } }`, else → `team: { key: { eq: $teamId } }`. Two named queries sharing the same response struct. The `len(nodes)==0` → `getWorkspaceLabelByName` fallback is untouched.

Closes #4965.

## Test plan
- [x] `go test ./internal/adapters/linear/...` — all 5 acceptance rows pass:
  - key-configured team, label exists → `team.key.eq`, no fallback
  - UUID-configured team, label exists → `team.id.eq`, no fallback
  - UUID-configured team, only workspace-scoped label → id-filter then workspace fallback
  - key-configured team, missing everywhere → key-filter then fallback, error
  - `GetOrCreateLabel` idempotency with UUID team + existing label → `CreateLabel` never called (the #4884 duplicate-create guard)
- [x] `go build ./...`
- [x] `go vet ./internal/adapters/linear/...`

### DIFF
--- a/internal/adapters/linear/client.go
+++ b/internal/adapters/linear/client.go
@@ -409,15 +409,33 @@ func (c *Client) RemoveLabel(ctx context.Context, issueID, labelID string) error
 	}, nil)
 }
 
-// GetLabelByName fetches a label ID by name for a team
+// GetLabelByName fetches a label ID by name for a team.
+// teamID may be a Linear team key (e.g. "ROU") or a team UUID — see
+// config.Config.TeamID ("Team ID or name"). Linear's `team.key` filter only
+// matches the short slug, so a UUID-configured team_id must be filtered via
+// `team.id` instead, or the lookup never matches (GH-4965, unfixed read half
+// of GH-4884; write half fixed in GH-4896's ResolveTeamUUID).
 func (c *Client) GetLabelByName(ctx context.Context, teamID, labelName string) (string, error) {
-	query := `
-		query GetLabel($teamId: String!, $name: String!) {
+	const queryByKey = `
+		query GetLabelByTeamKey($teamId: String!, $name: String!) {
 			issueLabels(filter: { team: { key: { eq: $teamId } }, name: { eq: $name } }) {
 				nodes { id name }
 			}
 		}
 	`
+	const queryByID = `
+		query GetLabelByTeamID($teamId: String!, $name: String!) {
+			issueLabels(filter: { team: { id: { eq: $teamId } }, name: { eq: $name } }) {
+				nodes { id name }
+			}
+		}
+	`
+
+	query := queryByKey
+	if looksLikeUUID(teamID) {
+		query = queryByID
+	}
+
 	var result struct {
 		IssueLabels struct {
 			Nodes []struct {
--- a/internal/adapters/linear/client_test.go
+++ b/internal/adapters/linear/client_test.go
@@ -1381,46 +1381,69 @@ func TestCreateLabelResolvesTeamKeyToUUID(t *testing.T) {
 }
 
 func TestGetLabelByNameWorkspaceFallback(t *testing.T) {
+	const teamUUID = "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
+
 	tests := []struct {
 		name          string
+		teamID        string
 		teamNodes     string
 		workspaceResp string
 		wantID        string
 		wantErr       bool
 		wantFallback  bool
+		wantFilter    string // "GetLabelByTeamKey" or "GetLabelByTeamID"
 	}{
 		{
-			name:         "team-scoped label found without fallback",
+			name:         "key-configured team, label exists",
+			teamID:       "ROU",
 			teamNodes:    `[{"id":"team-label","name":"llm-pilot"}]`,
 			wantID:       "team-label",
 			wantFallback: false,
+			wantFilter:   "GetLabelByTeamKey",
 		},
 		{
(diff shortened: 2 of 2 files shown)


alekspetrov:
## Post-merge review — VERDICT: APPROVE-with-defects ⚠️ CRITICAL

**The merged fix is a functional no-op for its target scenario.** The ID-branch query (`client.go:427`) declares `$teamId: String!` but uses it in `team: { id: { eq: $teamId } }`, where Linear's schema types `id` as `IDComparator` (`eq: ID`). GraphQL does not coerce `String!` into an `ID` position. **Empirically reproduced against api.linear.app**: the PR's exact `GetLabelByTeamID` document returns HTTP 400 `GRAPHQL_VALIDATION_FAILED: Variable "$teamId" of type "String!" used in position expecting type "ID"` — before auth is even checked. The same document with `$teamId: ID!` passes validation.

Consequence: for a UUID-configured `team_id`, `GetLabelByName` errors on every call → `GetOrCreateLabel` (client.go:589-597) falls through to `CreateLabel` → duplicate-name rejection → the exact #4884 symptom this PR was merged to fix.

**Fix is one token**: `$teamId: ID!` in `queryByID`. Follow-up issue: #TBD.

Secondary findings:
2. `client_test.go:1455-1461` — tests assert only the operation name in the request body; the httptest mock performs no GraphQL validation, which is precisely why this shipped green. The follow-up should at minimum pin the `$teamId: ID!` declaration in the query text.
3. Low: the "missing everywhere" test row checks `wantErr` only, not the spec'd `label %q not found in team %s` message.
4. Positive: composition with #4896 is clean — `looksLikeUUID` reused (strict 8-4-4-4-12, cannot collide with team keys), no duplicated resolution logic, key-configured path byte-identical to pre-PR.

PR #4981, opened 2026-08-18
GH-4968: fix(ghguard): honor GH_REPO/GH_HOST env in classification — env bypasses the repo check today

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-4968.

Closes #4968

## Changes

GitHub Issue GH-4968: fix(ghguard): honor GH_REPO/GH_HOST env in classification — env bypasses the repo check today

## Context

Spun out of TASK-479 research (open decision D5, `.agent/tasks/TASK-479-ghguard-parser-parity.md`): gh-guard classifies argv only. `gh` also reads `GH_REPO`/`GH_HOST` from the environment, so `GH_REPO=other/repo gh issue comment N --body x` comments on a foreign repo while the guard's repo check sees no `-R` flag and allows it — the exact cross-repo mutation class the guard exists to stop (GH-4649), and no argv-parser fix covers it.

## Problem

`runGhGuard` (cmd/pilot/ghguard.go) inherits the executor subprocess environment but never inspects it. The repo-scoping check in `internal/executor/ghguard/policy.go` (`checkOwnArtifact`) derives the target repo from `-R/--repo` argv only, defaulting to "current repo" when absent — while gh itself would honor `GH_REPO`.

## Implementation

- In `runGhGuard`, read `GH_REPO` and `GH_HOST` before classification. Treat a set `GH_REPO` exactly like an explicit `-R <value>` (feed it into the classification as the repo argument when argv has none; argv `-R` wins when both are present, matching gh's own precedence — verify precedence against gh docs/source and note it in a comment). A set `GH_HOST` pointing anywhere other than github.com ⇒ deny (fail closed — our repos live on github.com).
- Alternative considered and rejected: scrubbing the vars before exec'ing real gh would silently change the command's meaning; classify-with-env keeps the guard honest about what gh would actually do.
- Journal the env-derived repo in the denial audit record so a denied env-bypass attempt is distinguishable from an argv one.

## Acceptance

- `GH_REPO=other/repo gh issue comment 1 --body x` → deny (cross-repo mutation), same reason quality as the argv `-R` deny.
- `GH_REPO=<own repo> gh issue view 1` → allow.
- `GH_HOST=ghe.example.com gh api user` → deny.
- Argv `-R` + conflicting `GH_REPO`: argv wins (gh precedence), with a test row proving it.
- Table-driven tests alongside the existing policy tests; denial journal includes the env-derived repo.

## Refs

[...]

PR #4975, opened 2026-08-18
GH-4963: fix(ghguard): pflag-faithful per-command parser — close the 11 read/mutation parity gaps

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-4963.

Closes #4963

## Changes

GitHub Issue GH-4963: fix(ghguard): pflag-faithful per-command parser — close the 11 read/mutation parity gaps

no-decompose

Single coherent change in one subsystem (`internal/executor/ghguard/`) — do NOT decompose; one PR.

## Problem

`internal/executor/ghguard/` inspects `gh` command lines our autonomous executor is about to run and classifies each as read (allow) or mutation (deny). Correctness depends entirely on our `parseArgs` deriving the **same effective HTTP method and body-presence** that the real `gh` binary derives from the same argv. Where the two disagree, the guard is simply wrong.

`gh` uses pflag. Our parser models only space-separated flag values, one global flag table, and no bundling. pflag additionally accepts attached shorthand (`-XPOST` ≡ `-X POST`), `=` form (`-X=POST`), boolean bundles (`-piX POST`), last-occurrence-wins for scalars, and `--` termination — and `gh`'s flag grammar is **per-command** (`-f`, `-F`, `-p`, `-t`, `-s` mean different things under `api` vs `pr create` vs `issue comment`).

## Confirmed parity gaps (probed against main)

| # | argv | our parse | gh derives | today | correct |
|---|---|---|---|---|---|
| G1 | `api X -XPOST` | method="" | POST | allow | deny |
| G2 | `api X -XDELETE` | method="" | DELETE | allow | deny |
| G3 | `api X -fstate=closed` | hasData=false | POST+body | allow | deny |
| G4 | `api graphql -Fquery=x` | hasData=false | POST | allow | deny |
| G5 | `api X -pX DELETE` | method="" | DELETE | allow | deny |
| G6 | `api X -pXDELETE` | method="" | DELETE | allow | deny |
| G7 | `api X -p -X POST` | `-p` swallows `-X` | POST | allow | deny |
| G8 | `api X -X GET -XPOST` | GET | POST (last wins) | allow | deny |
| G9 | `issue view 1 -Rother/repo` | repoGiven=false | other repo | allow | deny |
| G10 | `issue comment N -Rother/repo --body x` | repoGiven=false | comments other repo | allow | deny |
| G11 | `pr create -f --head other-branch` | `-f` swallows `--head` | head=other-branch | allow | deny |

[...]

PR #4990, opened 2026-08-19
fix(ghguard): deny api graphql with data flags regardless of method (D2) + close --hostname argv twin of GH_HOST bypass (GH-4986)

### LINKED ISSUE #4986: feat(ghguard): deny api graphql with data flags regardless of method (D2) + close the --hostname argv twin of the GH_HOST bypass
## Problem

Two residual ghguard deny-gaps, both documented follow-up candidates from the TASK-479 reviews (PR #4975 APPROVE-with-notes, PR #4981 APPROVE-with-notes):

**D2 — `gh api graphql` with fields is never treated as a mutation.** GraphQL requests are POSTs whose read/write nature lives in the query text, not the HTTP method. Today a `gh api graphql -f query='mutation {...}'` classifies by method/data-flag heuristics that were designed for REST paths. The safe rule from the TASK-479 spec: deny `api graphql` whenever data-carrying flags (`-f`/`-F`/`--raw-field`/`--field`/`--input`) are present, regardless of method — a fields-carrying graphql call cannot be proven read-only from argv.

**`--hostname` — the argv twin of the GH_HOST env bypass #4981 just closed.** `internal/executor/ghguard/policy.go:280` registers `--hostname` as `{true, actionNone}`, so `gh api --hostname ghe.example.com user` is still allowed while `GH_HOST=ghe.example.com gh api user` is denied. The fail-closed rationale at policy.go:631-636 ("refusing to guess what host this call actually targets") applies verbatim to the argv form.

## Fix

1. In the per-command flag specs for `api`: when the subcommand target is `graphql`, deny if any data-carrying flag is present, regardless of `-X` method. Reason text should name the graphql+fields rule.
2. Change `--hostname` from `actionNone` to `actionSetHost` and run it through the same non-github.com deny added in #4981 (case-insensitive `github.com` no-op preserved).
3. Regression rows: `api graphql -f query=...` (deny) · `api graphql` bare (allow, read) · `api --hostname ghe.example.com user` (deny) · `api --hostname github.com user` (allow) · `--hostname` + GH_HOST both set (deny, host reason).

## Acceptance criteria

- [ ] `gh api graphql` with any of `-f/-F/--raw-field/--field/--input` denies regardless of method
- [ ] `--hostname <non-github.com>` denies exactly like GH_HOST; `github.com` (any case) is a no-op
- [ ] Journal entries distinguish the graphql-fields rule and hostname-flag rule from existing deny reasons
- [ ] Existing #4975/#4981 regression tables still green

## Refs

- TASK-479 doc (`.agent/tasks/TASK-479-ghguard-parser-parity.md`) D2 section
- https://github.com/qf-studio/pilot/pull/4975 (review notes) · https://github.com/qf-studio/pilot/pull/4981#issuecomment-5339806740 (finding 1)

Do not decompose — this is a standalone task, one coherent change to one package as a single PR.

alekspetrov:
## Summary

- `gh api graphql` now denies whenever a data-carrying flag (`-f`/`-F`/`--raw-field`/`--field`/`--input`) is present, regardless of `-X`/`--method` — GraphQL's read/write nature lives in the query text, not the HTTP verb, so the REST-shaped explicit-GET relaxation never applies to `api graphql` (D2, TASK-479).
- `--hostname` moves from `{true, actionNone}` to `actionSetHost`, flowing through the same fail-closed non-`github.com` deny that `GH_HOST` got in #4981 (case-insensitive `github.com` stays a no-op). Argv `--hostname` takes precedence over `GH_HOST` when both are set, mirroring `-R`/`--repo` over `GH_REPO`.
- Journal reasons distinguish the graphql-fields rule and the `--hostname` rule from existing deny reasons (e.g. doesn't mention `GH_HOST` when the deny came from argv `--hostname`).

Scope: `internal/executor/ghguard/` only, per the issue's "do not decompose" instruction.

Closes GH-4986.

## Test plan

- [x] `go test ./internal/executor/ghguard/...`
- [x] `go build ./...`
- [x] `go vet ./internal/executor/ghguard/...`
- [x] New regression rows: `api graphql -f query=...` (deny, explicit and default method) · `api graphql` bare (allow) · `api --hostname ghe.example.com user` (deny) · `api --hostname github.com`/case-insensitive (allow) · `--hostname` + `GH_HOST` both set, precedence both directions (deny/allow per issue spec)
- [x] Existing #4975/#4981 regression tables still pass

### DIFF
--- a/internal/executor/ghguard/policy.go
+++ b/internal/executor/ghguard/policy.go
@@ -32,23 +32,34 @@
 // mirrors pflag's per-command flag grammar — attached shorthand (-XPOST),
 // `=` forms, boolean bundling, last-occurrence-wins, and `--` termination —
 // via a per-(command,sub) flagSpec table (flagTableFor) rather than one
-// global flag list. Two decisions from the originating research were left
-// open rather than folded in here: denying `api graphql` with field flags
-// regardless of method (belt-and-braces; D2), and consuming known
-// value-flags while scanning for the sub when flags precede the
-// subcommand (a false-deny-only gap; D3).
+// global flag list. One decision from the originating research is still
+// left open rather than folded in here: consuming known value-flags while
+// scanning for the sub when flags precede the subcommand (a false-deny-only
+// gap; D3).
 //
-// GH_REPO/GH_HOST env bypass (GH-4968/D5): argv-only classification missed
-// that `gh` itself also honors GH_REPO/GH_HOST from the environment, so
-// `GH_REPO=other/repo gh issue comment N --body x` used to sail through the
-// -R/--repo check (which only ever looked at argv) straight into a
-// cross-repo mutation — the exact class GH-4649 exists to stop. Classify
-// now takes an EnvOverride carrying those two vars: a set GH_REPO is
-// treated exactly like an explicit -R/--repo when argv gave none (argv
-// still wins when both are present, matching gh's own -R > GH_REPO > local
-// git remote precedence — see EnvOverride's doc comment), and a non-
-// github.com GH_HOST is a hard, unconditional deny (fail closed: every
-// repo this guard reasons about lives on github.com).
+// api graphql with field flags (GH-4986/D2): GraphQL requests are POSTs
+// whose read/write nature lives in the query text, not the HTTP method, so
+// the REST-shaped explicit-GET relaxation in checkAPIRead never applies to
+// `gh api graphql` — any data-carrying flag (-f/-F/--raw-field/--field/
+// --input) denies regardless of -X/--method (belt-and-braces; a
+// fields-carrying graphql call cannot be proven read-only from argv).
+//
+// GH_REPO/GH_HOST env bypass (GH-4968/D5) and its --hostname argv twin
+// (GH-4986): argv-only classification missed that `gh` itself also honors
+// GH_REPO/GH_HOST from the environment, so `GH_REPO=other/repo gh issue
+// comment N --body x` used to sail through the -R/--repo check (which only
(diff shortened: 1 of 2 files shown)


alekspetrov:
## Pre-merge review — VERDICT: APPROVE (merging)

Both #4986 legs implemented with scoping that provably matches gh's own semantics: the graphql rule keys on `positional[0] == "graphql"` — the identical exact-match cli/cli uses (`opts.RequestPath == "graphql"`); path variants (`/graphql`, `GraphQL`, `graphql?x`) are REST paths to gh and covered by existing REST rules (fields + no explicit GET → deny; explicit GET → gh genuinely sends GET, which GitHub's POST-only GraphQL endpoint won't execute). No overreach: REST `-f ... -X GET` keeps the #4877/#4905 relaxation, existing #4975/#4981 regression tables pass untouched. `--hostname` is long-form-only in gh (no shorthand to bundle), attached `=` form handled, last-wins matches pflag, precedence over GH_HOST matches gh, journal reasons distinguish `--hostname` from `GH_HOST` denials. Full ghguard + cmd/pilot suites green on the PR head (97a31b66); vet clean.

One follow-up-grade nit (not blocking, not exploitable): `hostGiven` gates on flag *presence*, so `GH_HOST=ghe gh api --hostname= user` sets an empty effective host and skips the env deny — unreachable in practice because gh's own `HostnameValidator` rejects empty values before any request, and whitespace variants fail `EqualFold` → deny. One-line hardening: gate on `p.hostGiven && p.host != ""`.

PR #4984, opened 2026-08-18
GH-4982: fix(autopilot): missed-train recovery fires before the scheduled tick on boot, then a later restart never recovers the genuinely-missed tick

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-4982.

Closes #4982

## Changes

GitHub Issue GH-4982: fix(autopilot): missed-train recovery fires before the scheduled tick on boot, then a later restart never recovers the genuinely-missed tick

## Context

Live evidence 2026-08-18 (founder box, `~/.pilot/logs/daemon.log`): the pilot repo's daily train (`0 16 * * *` Europe/Berlin = 14:00Z) never ran for the day's post-13:45Z merges. Two cooperating defects:

1. **Early fire.** Daemon restart at 13:40:57Z (15:40 Berlin, i.e. BEFORE the 16:00 Berlin tick) ran boot-time missed-train recovery and cut **v2.261.0 at 13:45Z** — 15 minutes before the tick it was 'recovering'. A recovery pass on boot must only consider ticks whose scheduled time is in the PAST.
2. **No recovery when actually missed.** The next restart at 15:12:16Z (17:12 Berlin, AFTER the missed 16:00 Berlin tick) started the pilot scheduler with `next_run=2026-08-19T16:00` and logged **no** `recovering missed train` line for pilot — while the Mon–Fri (`0 16 * * 1-5`) schedulers for other repos at 15:11:27Z/15:11:40Z DID log recovery. Presumably the 13:45Z tag satisfied whatever freshness check gates recovery, even though 13:45Z predates the 14:00Z scheduled time.

Net effect: 15+ merges (sdk v0.35.1 pin #4954, KAN-6 #4955, #4971/#4973/#4975/#4977–#4981, CI fix 049456d5) sat unreleased with next_run a full day out. Operator cut v2.262.0 manually.

Log lines (UTC):
```
13:40:57 platform-outage breaker enabled            (daemon boot #1)
13:45:07 tag v2.261.0 created                        (git tag creatordate; 15min BEFORE tick)
15:11:27 release train scheduler started schedule="0 16 * * 1-5" next_run=2026-08-19T16:00
15:11:27 recovering missed train scheduled_at=2026-08-18T16:00:00+02:00
15:11:28 scheduleReleaseTick: no resolvable member PRs (direct-commit-only train), skipping — v1 limitation last_tag=v0.1.0 commits=9
[...]

Issue #4982, opened 2026-08-18
fix(autopilot): missed-train recovery fires before the scheduled tick on boot, then a later restart never recovers the genuinely-missed tick


alekspetrov:
## Context

Live evidence 2026-08-18 (founder box, `~/.pilot/logs/daemon.log`): the pilot repo's daily train (`0 16 * * *` Europe/Berlin = 14:00Z) never ran for the day's post-13:45Z merges. Two cooperating defects:

1. **Early fire.** Daemon restart at 13:40:57Z (15:40 Berlin, i.e. BEFORE the 16:00 Berlin tick) ran boot-time missed-train recovery and cut **v2.261.0 at 13:45Z** — 15 minutes before the tick it was 'recovering'. A recovery pass on boot must only consider ticks whose scheduled time is in the PAST.
2. **No recovery when actually missed.** The next restart at 15:12:16Z (17:12 Berlin, AFTER the missed 16:00 Berlin tick) started the pilot scheduler with `next_run=2026-08-19T16:00` and logged **no** `recovering missed train` line for pilot — while the Mon–Fri (`0 16 * * 1-5`) schedulers for other repos at 15:11:27Z/15:11:40Z DID log recovery. Presumably the 13:45Z tag satisfied whatever freshness check gates recovery, even though 13:45Z predates the 14:00Z scheduled time.

Net effect: 15+ merges (sdk v0.35.1 pin #4954, KAN-6 #4955, #4971/#4973/#4975/#4977–#4981, CI fix 049456d5) sat unreleased with next_run a full day out. Operator cut v2.262.0 manually.

Log lines (UTC):
```
13:40:57 platform-outage breaker enabled            (daemon boot #1)
13:45:07 tag v2.261.0 created                        (git tag creatordate; 15min BEFORE tick)
15:11:27 release train scheduler started schedule="0 16 * * 1-5" next_run=2026-08-19T16:00
15:11:27 recovering missed train scheduled_at=2026-08-18T16:00:00+02:00
15:11:28 scheduleReleaseTick: no resolvable member PRs (direct-commit-only train), skipping — v1 limitation last_tag=v0.1.0 commits=9
15:12:16 release train scheduler started schedule="0 16 * * *" timezone=Europe/Berlin next_run=2026-08-19T16:00:00.000+02:00
         (no recovery line for the pilot scheduler despite the 14:00Z tick being missed)
```

## Acceptance

[...]

Issue #4989, opened 2026-08-19
fix(autopilot): train-recovery logging announces before deciding and skips silently — the exact gap behind the #4982 misdiagnosis


alekspetrov:
## Problem

Post-merge review of PR #4984 (GH-4982) + box-log verification exposed a diagnosability gap that directly caused the #4982 misdiagnosis ("defect #2" — the 14:00Z tick was never actually missed; the row-exists skip was correct but invisible):

1. `recovering missed train` (WARN) logs **before any gate is evaluated** — it fired at every daemon boot on 08-17/08-18 (16:49, 18:22, 18:55, 12:02, 12:07, 12:53, 13:41…) including boots where recovery then did nothing. The log announces an action that mostly doesn't happen.
2. The row-exists return (`internal/autopilot/scope_schedule.go:246-250`) logs **nothing** — the one skip that mattered on 08-18 15:11Z was silent, which is why the incident was misread as "recovery never ran".
3. Skip verdicts log at Debug (`:233` lookback, `:267` last-release gate) — invisible at the daemon's Info level.

## Fix

- Move the `recovering missed train` WARN to AFTER all gates pass (i.e., log it only when a recovery release will actually be attempted).
- Add an Info-level log to the row-exists return: repo, scope key, row state, row tag.
- Promote the lookback and last-release skip verdicts from Debug to Info, each with scheduled_at + the gate that decided.

One Info line per boot per repo describing the recovery decision (fire, or skip + which gate) satisfies the #4982 acceptance row "diagnosable from one log line".

## Acceptance criteria

- [ ] No `recovering missed train` WARN unless a recovery release is actually attempted
- [ ] Every skip path (lookback, row-exists, last-release, error) emits one Info line with the reason
- [ ] Existing #4984 tests still green; new test asserts log side of each verdict

## Refs

https://github.com/qf-studio/pilot/pull/4984#issuecomment-5339816401 (review finding 2) and the box-verification comment on the same PR.

Do not decompose — this is a standalone task, one small change as a single PR.

alekspetrov:
🤖 **Pilot started working on this issue**

Task ID: `GH-4989`

I'll post updates as I make progress.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #4993 |
| Branch | `pilot/GH-4989` |
| Time to merge | 8m45s |

PR #4993, opened 2026-08-19
fix(autopilot): train-recovery WARN fires only after gates pass, skips log at Info

alekspetrov:
## Summary
- The "recovering missed train" WARN in `recoverMissedTrainTick` was already structurally gated behind `decideTrainRecovery`'s Fire verdict, but the surrounding skip paths were not diagnosable: the row-exists skip (the one that mattered in the #4982 misdiagnosis) logged nothing, and the lookback / last-release skip verdicts logged at Debug — invisible at the daemon's Info level.
- Promotes all three skip paths (lookback, row-exists, last-release) to Info, each naming the deciding gate (`gate=lookback|row-exists|last-release`) plus `repo`/`scheduled_at`/reason context. The row-exists skip additionally logs `row_state`/`row_tag`.
- Error paths (`row-lookup-error`, `last-release-lookup-error`) keep their existing Warn level, now also tagged with `gate=` and `repo=` for consistency.

## Why
Closes the diagnosability gap from the #4982 post-merge review (PR #4984 comment) and box-log verification: a correct, silent row-exists skip on 2026-08-18 15:11Z was misread as "recovery never ran" because nothing logged the reason. One Info line per boot per repo now states fire-or-skip-and-why.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/autopilot/...` (full package, includes existing #4984 `TestRecoverMissedTrainTick*` / `TestDecideTrainRecovery` — all green)
- [x] New `TestRecoverMissedTrainTick_LogsOneLinePerVerdict` (gh4989_test.go) asserts the log side of all four verdicts: lookback skip, row-exists skip, last-release skip, and the fire path (WARN appears exactly once, never on a skip)
- [x] `go test ./...` (full repo suite green)

### DIFF
--- a/.agent/tasks/gh-4989.md
+++ b/.agent/tasks/gh-4989.md
@@ -1,0 +1,38 @@
+# GH-4989
+
+**Created:** 2026-08-19
+
+## Problem
+
+GitHub Issue GH-4989: fix(autopilot): train-recovery logging announces before deciding and skips silently — the exact gap behind the #4982 misdiagnosis
+
+## Problem
+
+Post-merge review of PR #4984 (GH-4982) + box-log verification exposed a diagnosability gap that directly caused the #4982 misdiagnosis ("defect #2" — the 14:00Z tick was never actually missed; the row-exists skip was correct but invisible):
+
+1. `recovering missed train` (WARN) logs **before any gate is evaluated** — it fired at every daemon boot on 08-17/08-18 (16:49, 18:22, 18:55, 12:02, 12:07, 12:53, 13:41…) including boots where recovery then did nothing. The log announces an action that mostly doesn't happen.
+2. The row-exists return (`internal/autopilot/scope_schedule.go:246-250`) logs **nothing** — the one skip that mattered on 08-18 15:11Z was silent, which is why the incident was misread as "recovery never ran".
+3. Skip verdicts log at Debug (`:233` lookback, `:267` last-release gate) — invisible at the daemon's Info level.
+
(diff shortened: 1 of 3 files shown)


alekspetrov:
## Post-merge review — VERDICT: APPROVE-with-notes

Surgical fix: all three silent/Debug skip paths now emit one Info line naming the deciding gate (`gate=lookback|row-exists|last-release`) with repo and timing context; error skips stay at Warn (louder — satisfies the intent) and gain gate tags. `decideTrainRecovery` has zero diff lines — gate ordering and #4984 semantics untouched. The new `TestRecoverMissedTrainTick_LogsOneLinePerVerdict` captures real slog output at Debug level, so the Info promotion is genuinely asserted (a Debug-level implementation would fail it), and the fire-path WARN is pinned exactly-once.

**Premise correction, handled correctly by the PR**: issue #4989's requirement 1 (move the WARN behind the gates) was already satisfied on main — PR#4984's `decision.Fire` check precedes the WARN, and the every-boot WARNs in the 08-17/18 box logs came from pre-#4984 builds. The PR documents this instead of churning code, and pins the contract with a test. Right call.

Nits: the row-exists Info line encodes `scheduled_at` inside `scope=` rather than as a named field (grep by `scheduled_at=` misses that one path); `.agent/tasks/gh-4989.md` has a duplicated header + empty AC section and should be archived now that this is merged; `downstreamHits` in the last-release subtest is captured but never asserted.

PR #5016, opened 2026-08-20
feat(executor,memory): Contract Evidence gate executor-side implementation (GH-5012)

alekspetrov:
## Summary

Executor-side leg of the Contract Evidence enforcement gate (TASK-460 doc-vs-wire leg, part of #5009). Implements items (a)-(g) from #5012's scope:

- `internal/executor/contract_evidence.go` — executor-local `ContractDependency` mirror (no `internal/config` import, avoids an import cycle), `--json-schema` constant, `detectTouchedContractFields` (recall-oriented Go `json:"..."` tag / TS interface field scan over glob-matched diff hunks), and `verifyContractEvidence` enforcing all four rejection rules from Requirement 5: field-not-in-diff, citation's repo not a configured dependency, cited producer line's ±3-line window doesn't contain the field/`ProducingExpr`, and missing-for-required. Fetch errors (including no fetcher configured) are hard failures, never a silent pass.
- `internal/executor/runner.go` — `SetContractDependencyLookup`/`SetContractContentFetcher`, `getContractEvidence` (mirrors `getPostExecutionSummary`), and the gate spliced between self-review/intent-judge and the `DirectCommit`/`CreatePR` branch — identical failure shape to the existing `QualityChecker` gate (`result.Success=false`, `AlertEventTypeTaskFailed`, `webhooks.EventTaskFailed`, `recorder.Finish("failed")`). No-op (zero new GitHub API calls) when no lookup is configured or a project declares no contract dependencies.
- `internal/executor/prompt_builder.go` — advisory-only Contract Evidence section in `buildSelfReviewPrompt` (explicitly documented as NOT the enforcement mechanism — the gate independently re-derives and verifies evidence) plus `buildContractEvidencePrompt` for the gate's own structured-output fetch.
- `internal/memory/store.go` — `StageContractEvidence` event stage, following the `StageQualityGate` doc-comment convention.
- `contract_evidence_test.go` / `runner_contract_evidence_test.go` — unit and runner-level integration coverage for every acceptance path: no-config/no-dependency no-ops, zero-evidence failure, irrelevant-citation failure, wrong-producer-line failure, valid-citation success, alert/webhook/recorder-finish assertions via real (not mocked) sinks, and a synthetic fixture reproducing the ui PR#113 consumer-docblock-citation incident shape.

[...]

Issue #5009, opened 2026-08-20
Contract Evidence enforcement gate: block false wire-contract delivery via producer-source verification (TASK-460 doc-vs-wire leg)


alekspetrov:
## Context

TASK-460 has recurred 4 times: the executor ships a wire-contract change with green CI that doesn't deliver the requested semantics, because it trusted a stale same-repo docblock instead of the actual producer source. Most recent instance: ui PR#113 consumed `specVersion` from `pilot-console-ui/src/lib/api/types.ts`, trusting a docblock claiming it was the APPLIED generation, when the producer (`pilot-console/internal/instances/handlers.go:141`) sends the TARGET generation. Tests enshrined the wrong semantics and CI stayed green.

A prompt-only fix (another self-review checklist item) was rejected as cosmetic — self-review output is only ever advisory-grepped (`strings.Contains(result.Output, "REVIEW_PASSED")` in `internal/executor/runner.go`), never a hard gate. This issue adds a genuinely enforced, machine-validated gate: when a project declares wire-contract dependencies, the executor must cite real producer source for fields it asserts semantics about, and the daemon independently fetches and verifies those citations via the GitHub API before allowing task success.

Key existing seams (verified):

- Hard-block precedent: the `QualityChecker` loop at `internal/executor/runner.go:4344-4394` (sets `result.Success=false`, emits `AlertEventTypeTaskFailed`, dispatches `webhooks.EventTaskFailed`, `recorder.Finish("failed")`, returns). The new gate must fail through this exact shape.
- Structured-output precedent: `internal/executor/structured_output.go` (`PostExecutionSummarySchema` + `extractStructuredOutput`), consumed by `getPostExecutionSummary` (`runner.go:6731`). Reuse this mechanism — do NOT parse evidence out of freeform self-review text.
- Per-project wiring precedent: `internal/executor` has zero import of `internal/config` (avoid the cycle). Follow `newProjectQualityCheckerFactory` (`cmd/pilot/adapters.go:199`) → `runner.SetQualityCheckerFactory(...)` at 5 call sites in `main.go`/`commands.go`.
- GitHub client (`internal/adapters/github/client.go`) has no Contents API method; a new one inherits `internal/ghbudget` rate-budget protection for free (Transport is nil → budget RoundTripper applies).
- Insertion point: after the self-review/intent-judge `wg.Wait()` (`runner.go:4749`) and before the `task.DirectCommit`/`task.CreatePR` branch (`runner.go:4851`).

[...]

PR #5017, opened 2026-08-20
GH-5013: feat(cmd/pilot): wire ContractDependencyLookup into all runner construction sites

### LINKED ISSUE #5013: feat(cmd/pilot): wire ContractDependencyLookup into all runner construction s...
Parent: GH-5009

Add newProjectContractDependencyLookup(cfg *config.Config) in cmd/pilot/adapters.go (mirrors newProjectQualityCheckerFactory at :199), returning an executor-local ContractDependency slice per project path (bridging internal/config → internal/executor without an import cycle). Call runner.SetContractDependencyLookup(...) at every one of the 5 SetQualityCheckerFactory call sites across cmd/pilot/main.go and cmd/pilot/commands.go. Update cmd/pilot/wiring_test.go to assert the lookup is set on every runner instantiation path — the fail-when-unwired guardrail that prevents this from silently no-op'ing in production. Depends on subtasks 1 and 3.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5009 is decomposed

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5013.

Closes #5013

## Changes

Add newProjectContractDependencyLookup(cfg *config.Config) in cmd/pilot/adapters.go (mirrors newProjectQualityCheckerFactory at :199), returning an executor-local ContractDependency slice per project path (bridging internal/config → internal/executor without an import cycle). Call runner.SetContractDependencyLookup(...) at every one of the 5 SetQualityCheckerFactory call sites across cmd/pilot/main.go and cmd/pilot/commands.go. Update cmd/pilot/wiring_test.go to assert the lookup is set on every runner instantiation path — the fail-when-unwired guardrail that prevents this from silently no-op'ing in production. Depends on subtasks 1 and 3.


alekspetrov:
**Review verdict: APPROVE, conditional** — merge AFTER #5016, and the activation follow-up issue must be filed first (it now is — see below).

Verified against current main: exactly 5 production SetQualityCheckerFactory sites exist (main.go:734/1085/1247/2089, commands.go:761) and this PR pairs every one with SetContractDependencyLookup, including the webhook-mode path via the new Pilot→Orchestrator bridge methods. The parity tripwire in wiring_test.go is the right guardrail — it pins the baseline count, requires parity, and checks per-receiver pairing, so a sixth runner-construction site can't silently ship without the gate. Condition: nothing in this chain wires SetContractContentFetcher (the PR's own FEATURE-MATRIX entry says so), which means any project that sets contract_dependencies will hard-fail every contract-touching task with fetch_error — fail-closed, but unusable, and #5009 is already closed so it was untracked. Sequencing: this branch contains all of #5016's diff — merge #5016 first so this PR reduces to the wiring delta.

Issue #5019, opened 2026-08-20
Contract Evidence activation: wire SetContractContentFetcher at all runner sites + close review defects (docblock, diff-error fail-open, empty-fields guard)


alekspetrov:
## Context

Post-merge/pre-merge review of the #5009 Contract Evidence chain (PR#5014/#5015 merged, PR#5016/#5017 approved) found a decomposition gap plus three small defects. This issue is the activation leg — until it ships, no project may set `contract_dependencies`: the gate would hard-fail every contract-touching task with `fetch_error` (fail-closed lockout, honestly documented in the FEATURE-MATRIX entry from PR#5017).

## Requirements

1. **Wire the content fetcher (the blocker).** `Runner.SetContractContentFetcher` (defined in PR#5016) is never called anywhere in `cmd/pilot`. Construct the fetcher from the existing `*github.Client` (its `GetFileContent` from PR#5015) and inject it at every site `SetContractDependencyLookup` is called (5 sites: `cmd/pilot/main.go` ×4, `cmd/pilot/commands.go` ×1), including the `Pilot`→`Orchestrator` bridge methods added in PR#5017. Extend the wiring-parity tripwire test (`cmd/pilot/wiring_test.go`, `TestContractDependencyLookupWiredAtEveryQualityCheckerFactorySite` pattern) to also require `SetContractContentFetcher` parity.
2. **Fix the `ContractFiles` docblock** (`internal/config/contract_dependencies.go:15-18`) — it currently describes producer-repo paths, but the executor matches these globs against the CONSUMING project's diff as the gate trigger. Reword to: "glob allowlist of paths in the consuming project whose changes trigger the gate". Also fix the misleading test fixture (`contract_dependencies_test.go:98` uses a producer-shaped `internal/gateway/api.go` path). A user following today's docblock configures a gate that silently never fires — the misconfiguration twin of the incident class.
3. **Close the diff-error fail-open crack** (`internal/executor/runner.go` ≈:4906, from PR#5016): when the lookup is configured and the project has contract dependencies, a `GetDiffAgainstOrigin` error must FAIL the task through the standard contract-evidence failure sequence, not skip the gate with a Warn log. (Skip remains correct when lookup unset / no deps.)
4. **Guard the empty-fields LLM call**: when `required=true` but zero field tokens were extracted, skip `getContractEvidence` (no LLM call for an empty field list) and treat as passed with `Required=false` recorded consistently.

## Acceptance Criteria

[...]

PR #5026, opened 2026-08-20
GH-5021: fix(executor): close diff-error fail-open crack and guard empty-fields LLM call

### LINKED ISSUE #5021: fix(executor): close diff-error fail-open crack and guard empty-fields LLM call
Parent: GH-5019

Two related defects in the same package. (a) In `internal/executor/runner.go` (~line 4906, from PR#5016), when the contract-dependency lookup is configured AND the project has contract dependencies, a `GetDiffAgainstOrigin` error must route through the standard contract-evidence failure sequence (alert + webhook + recorder) with `result.Success=false`, not skip the gate with a Warn log; preserve the current skip-and-warn behavior when lookup is unset or the project has no deps. (b) In `internal/executor/contract_evidence.go`, when `required=true` but zero field tokens were extracted, short-circuit before `getContractEvidence` (no LLM/structured-output subprocess call) and record the result as passed with `Required=false` consistently. Add tests: configured-project diff-error → contract-evidence failure path fires (assert alert + webhook + recorder); unconfigured project + same failure → unchanged; zero-extracted-fields → call-count assertion proves no LLM invocation and task proceeds. Both defects live in `internal/executor/` — bundled to avoid intra-package merge conflicts.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5019 is decomposed

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5021.

Closes #5021

## Changes

GitHub Issue GH-5021: fix(executor): close diff-error fail-open crack and guard empty-fields LLM call

Parent: GH-5019

Two related defects in the same package. (a) In `internal/executor/runner.go` (~line 4906, from PR#5016), when the contract-dependency lookup is configured AND the project has contract dependencies, a `GetDiffAgainstOrigin` error must route through the standard contract-evidence failure sequence (alert + webhook + recorder) with `result.Success=false`, not skip the gate with a Warn log; preserve the current skip-and-warn behavior when lookup is unset or the project has no deps. (b) In `internal/executor/contract_evidence.go`, when `required=true` but zero field tokens were extracted, short-circuit before `getContractEvidence` (no LLM/structured-output subprocess call) and record the result as passed with `Required=false` consistently. Add tests: configured-project diff-error → contract-evidence failure path fires (assert alert + webhook + recorder); unconfigured project + same failure → unchanged; zero-extracted-fields → call-count assertion proves no LLM invocation and task proceeds. Both defects live in `internal/executor/` — bundled to avoid intra-package merge conflicts.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5019 is decomposed

### DIFF
--- a/internal/executor/contract_evidence.go
+++ b/internal/executor/contract_evidence.go
@@ -268,6 +268,42 @@ func matchesAnyContractGlob(path string, deps []ContractDependency) bool {
 	return false
 }
 
+// shortCircuitEmptyContractFields decides, from detectTouchedContractFields'
+// required/fields output, whether the getContractEvidence LLM/structured-
+// output subprocess call is needed at all (GH-5021b).
+//
+// A contract file can be touched (required=true) by a hunk that adds no
+// line matching either field-token regex — e.g. a comment-only or
+// non-field change inside an allow-listed file. There is nothing to cite in
+// that case, so calling out to getContractEvidence would spend a real LLM
+// invocation on an empty field set for no reason. This short-circuits
+// before that call: it builds the outcome directly via
+// verifyContractEvidence's existing empty-fields behavior (nil evidence,
+// empty requiredFields), which reports Required=false/Passed=true —
+// deliberately reusing that path rather than hand-building an equivalent
+// trivial outcome, so the two "nothing to verify" cases (no contract file
+// touched at all vs. touched-but-fieldless) stay consistent.
+//
+// Returns a non-nil outcome (and needsLLM=false) when the caller should
+// skip getContractEvidence and use the outcome as-is; returns a nil outcome
+// with needsLLM=true when the caller must still call getContractEvidence
+// and run its result through verifyContractEvidence itself.
+func shortCircuitEmptyContractFields(
+	ctx context.Context,
+	fetcher ContractContentFetcher,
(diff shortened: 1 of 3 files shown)


alekspetrov:
**Review verdict: APPROVE** (post-merge, GH-5021 re-land)

Both halves deliver the reviewed intent. The fail-open crack is closed correctly — deps resolution hoisted above the diff call (safe: the whole gate is guarded by `contractDependencyLookup != nil` at runner.go:4897), and the extraction of `failContractEvidenceGate` from the existing rejected-citation path means the failure sequence has one definition instead of two, which is better than the spec asked for. The fail-path test earns its keep: real repo pair, genuinely induced diff failure, and all three legs (alert, webhook, recorder metadata.json status=failed) asserted via fakes rather than log-grepping. The zero-fields guard reuses `verifyContractEvidence`'s empty-path (`Required=false/Passed=true`, contract_evidence.go:335-339) and the `fetchCalls==0` assertion goes through the actual `getContractEvidence` override seam, so it proves what it claims. One operational note: with deps configured, ANY diff-compute error now hard-fails the task even if the change touched no contract file — intended (fail closed), but keep it in mind for flaky-git triage.

Issue #5032, opened 2026-08-20
fix(autopilot): ensure parked PR resumes via existing re-drive path


alekspetrov:
Parent: GH-5027

Locate the mechanism that un-parks PRs held for base-mismatch today and confirm/mirror it so that a PR parked for the stacked-superset reason resumes through the same re-drive path once the blocking PR merges.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5027 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

## Review-discovered scope additions (2026-08-20 post-merge review of PR#5037/PR#5040 — REQUIRED, not optional)

The merged park implementation has NO exit half. Post-merge review findings to deliver in this issue:

1. **Un-park cleanup, mirroring the GH-4911 idiom** (`parkForBaseMismatch`'s cleanup at `controller.go:4401-4412`): when a stacked-parked PR resumes, clear `Parked`, clear the `stackedSupersetReasonPrefix` escalation reason, and remove the `autopilot/parked-awaiting-approval` label from the issue. Today `stackedSupersetReasonPrefix` appears nowhere outside `parkForStackedSuperset` — the resumed PR merges with `Parked=true` and a stale persisted reason (both survive restarts per GH-4598), and the label is never removed.
2. **Resume on base MERGE, not base finalization**: the descendant currently un-parks only when the base PR leaves `activePRs` (after post-merge CI + release). That is resume latency normally and a PERMANENT park if the base wedges tracked in a terminal stage post-merge. Un-park when the base reaches merged (or equivalently: filter `detectStackedSuperset` candidates by stage so merged/StageFailed PRs don't count as stack bases — pick one, document why).
3. **Assertions PR#5040 deliberately left to this issue**: extend `TestController_..._UnparksAfterBaseMerges` (controller_gh5034 tests) so tick 2 asserts `Parked==false` AND parked-label removal — both would fail against today's main; that is the fail-when-unwired proof this issue needs. Also model the base's merge as a stage transition (StageMerged, still tracked), not `delete(activePRs, ...)`, to pin resume-on-merge.
4. Candidate-stage filtering in `detectStackedSuperset` (`controller.go:6429`): exclude merged-but-still-tracked and StageFailed PRs from the "stacked on" candidate set (noted in PR#5035 review).

alekspetrov:
[...]

PR #5035, opened 2026-08-20
GH-5029: feat(autopilot): implement detectStackedSuperset ancestry helper

### LINKED ISSUE #5029: feat(autopilot): implement detectStackedSuperset ancestry helper
Parent: GH-5027

In internal/autopilot/controller.go, add a private helper (e.g. detectStackedSuperset) that, for a PR whose base is the default branch, checks each other open autopilot PR in the same repo to determine whether the current PR's head is a strict descendant of that PR's head. Use `git merge-base --is-ancestor` in a local worktree (reusing primitives from attemptLocalMerge / conflict_worktree_test.go) as the primary detection path, with the GitHub compare API as a fallback when a local clone isn't available. Document the choice of local-git-first inline.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5027 is decomposed

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5029.

Closes #5029

## Changes

In internal/autopilot/controller.go, add a private helper (e.g. detectStackedSuperset) that, for a PR whose base is the default branch, checks each other open autopilot PR in the same repo to determine whether the current PR's head is a strict descendant of that PR's head. Use `git merge-base --is-ancestor` in a local worktree (reusing primitives from attemptLocalMerge / conflict_worktree_test.go) as the primary detection path, with the GitHub compare API as a fallback when a local clone isn't available. Document the choice of local-git-first inline.

### DIFF
--- a/internal/autopilot/conflict_worktree.go
+++ b/internal/autopilot/conflict_worktree.go
@@ -2,6 +2,7 @@ package autopilot
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


alekspetrov:
**Review verdict: APPROVE-w-notes** (post-merge)

Detection direction, strictness, and the ancestor-never-blocked requirement are all correct and tested from both sides; the fail-through-to-API (rather than silent "not stacked") on local-git failure is the right call. Two notes for a fast-follow: the doc comment's "two git fetches total" undersells the real cost — it's one fetch plus a scratch-worktree add/remove per candidate per tick, and the worktree is unnecessary since `merge-base --is-ancestor` only reads the object DB (run it in `repoPath` directly). Also consider filtering candidates by stage: a merged-but-still-tracked or StageFailed PR shouldn't count as something to be "stacked on" — folded into GH-5032's scope (item 4 of its review additions).

Issue #5044, opened 2026-08-20
feat(github): add base-presence probe helpers


alekspetrov:
Parent: GH-5028

In internal/adapters/github/, add two thin, testable helpers on the existing client: FileExistsOnDefaultBranch(ctx, owner, repo, path) (bool, error) (wraps GetContents on the repo's default branch; treats 404 as false, nil) and IssueOrPRState(ctx, owner, repo, number) (kind, state string, err error) (returns whether #N is an issue with an open/merged/closed PR attached, or an open/merged PR itself — reuses go-github's pulls/issues endpoints). Ship with httptest-driven unit tests covering the 404-as-absent, open-PR, and merged-PR shapes. Nothing else calls these yet, so no wiring beyond the client file + its _test.go. This exists purely so subtask 2 can consume it without cross-package churn.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5028 is decomposed
into sibling sub-issues that cover the rest of its spec — consult the parent
for context, but do NOT implement parts of it that fall outside this subtask.

alekspetrov:
🤖 **Pilot started working on this issue**

Task ID: `GH-5044`

I'll post updates as I make progress.

alekspetrov:
⏳ Work delivered via https://github.com/qf-studio/pilot/pull/5046 but PR is still open — issue stays open until it merges.

alekspetrov:
✅ PR merged successfully!

| Metric | Value |
|--------|-------|
| PR | #5046 |
| Branch | `pilot/GH-5044` |
| Time to merge | 9m10s |

Issue #5028, opened 2026-08-20
Dispatcher: base-presence check before claim — hold issues whose prerequisite PR/files are not yet on main


alekspetrov:
## Context

Named daemon-side candidate from the 2026-08-15 incident wave (pitfall `sequential-gates-on-execution-not-merge-fastfollow-misbase`, backlog line in `.agent/DEVELOPMENT-README.md:221`), never implemented — and its absence produced fresh toil on 2026-08-20: three branches (pilot PR#5024, ui PR#126/#130/#141-class) were cut before their prerequisite PRs merged, went CONFLICTING, failed autopilot, and needed operator close-to-rearm cycles.

The pattern: an issue is claimed and executed while the PR it builds on is still open; the executor branches from a main that lacks the prerequisite content, then either vendors it (scope-fence break), stacks on the unmerged branch (see sibling ancestry-check issue), or produces a conflicting diff.

## Requirements

1. Before the dispatcher/claim path executes an issue, run a base-presence check: if the issue body references files, symbols, or a prerequisite PR/issue (`Depends on: #N`, "after #N merges", explicit file paths) whose content is not yet on the default branch, HOLD dispatch instead of claiming.
2. Concrete signals, cheapest-first (implement at least a+b):
   a. Issue body contains `Depends on: #N` / `Blocked by: #N` where #N is an issue whose PR is open-unmerged, or is an open PR → hold. (Reuse/extend the reference regex from `internal/executor/dependency_detector.go:34` — but at DISPATCH time, against live GitHub state, not execution time.)
   b. Issue body names a file path (backtick-quoted path heuristic) that does not exist on the default branch → hold with a "prerequisite not on main?" reason.
3. Hold behavior: do not claim; re-check on subsequent poll cycles (natural backoff); after a bounded number of held cycles (configurable, default generous), escalate with `pilot-needs-human` instead of holding forever — a genuinely-missing file reference must not silently starve the issue.
4. Observability: held-for-base-presence must be visible (log line + ledger/journal reason), not a silent skip — per the `poller-labels-removed-log-means-never-applied` lesson, absence of action with no signal reads as a wedge.
5. No behavior change for issues with no dependency refs and no missing-path hits (the overwhelming default) — zero extra GitHub API calls beyond what the poller already fetched, except the file-existence probes when a path is actually named.

## Acceptance Criteria

[...]

PR #5047, opened 2026-08-20
GH-5045: feat(executor): dispatcher base-presence gate before claim

### LINKED ISSUE #5045: feat(executor): dispatcher base-presence gate before claim
Parent: GH-5028

All work in internal/executor/. (a) Extend dependency_detector.go with two exported extractors reused at dispatch time: ExtractDependencyRefs(body string) []int (generalizes the existing dependencyRefRe — no sibling scoping, since at dispatch we don't have a sibling set) and ExtractReferencedPaths(body string) []string (backtick-quoted path heuristic with a small ignore-list for URLs). (b) Introduce a basePresenceChecker with a probe-interface satisfied by the subtask-1 helpers (kept as an interface so tests inject fakes; the concrete adapter wiring lives at the dispatcher construction seam already used for other GitHub access). (c) In the claim path in dispatcher.go, before the label pickup commits to executing: if the issue body yields any refs or paths, run the probes; on any open-PR ref or missing-path hit, do NOT claim — record a hold in the existing ledger/journal with a distinct reason string (held: prerequisite not on main), increment a per-issue hold counter (new memory-store column or reuse the existing backoff row shape near RepickBackoffState / ClaimLostDropCount), and emit a structured log line at info level. (d) After a bounded number of held cycles (add DispatcherConfig.BasePresenceHoldMaxCycles with a generous default), escalate by applying the pilot-needs-human label via the existing GitHub adapter path and clear the hold counter. (e) When no refs and no paths are extracted, the call path must be byte-identical to today — enforce with a call-count assertion in tests (zero probe calls). Tests: table-driven extractor tests using real bodies from GH-5021, ui GH-120/124/139 as fixtures; a claim-path integration test with a fake tracker/probe that flips PR state from open→merged across two poll ticks and asserts claim only on the second tick; an escalation test that runs N+1 cycles and asserts pilot-needs-human was applied and the hold cleared; and a 'no refs, no paths' fast-path test asserting zero probe calls.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5028 is decomposed

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5045.

Closes #5045

## Changes

All work in internal/executor/. (a) Extend dependency_detector.go with two exported extractors reused at dispatch time: ExtractDependencyRefs(body string) []int (generalizes the existing dependencyRefRe — no sibling scoping, since at dispatch we don't have a sibling set) and ExtractReferencedPaths(body string) []string (backtick-quoted path heuristic with a small ignore-list for URLs). (b) Introduce a basePresenceChecker with a probe-interface satisfied by the subtask-1 helpers (kept as an interface so tests inject fakes; the concrete adapter wiring lives at the dispatcher construction seam already used for other GitHub access). (c) In the claim path in dispatcher.go, before the label pickup commits to executing: if the issue body yields any refs or paths, run the probes; on any open-PR ref or missing-path hit, do NOT claim — record a hold in the existing ledger/journal with a distinct reason string (held: prerequisite not on main), increment a per-issue hold counter (new memory-store column or reuse the existing backoff row shape near RepickBackoffState / ClaimLostDropCount), and emit a structured log line at info level. (d) After a bounded number of held cycles (add DispatcherConfig.BasePresenceHoldMaxCycles with a generous default), escalate by applying the pilot-needs-human label via the existing GitHub adapter path and clear the hold counter. (e) When no refs and no paths are extracted, the call path must be byte-identical to today — enforce with a call-count assertion in tests (zero probe calls). Tests: table-driven extractor tests using real bodies from GH-5021, ui GH-120/124/139 as fixtures; a claim-path integration test with a fake tracker/probe that flips PR state from open→merged across two poll ticks and asserts claim only on the second tick; an escalation test that runs N+1 cycles and asserts pilot-needs-human was applied and the hold cleared; and a 'no refs, no paths' fast-path test asserting zero probe calls.

### DIFF
--- a/internal/executor/base_presence.go
+++ b/internal/executor/base_presence.go
@@ -1,0 +1,277 @@
+// Package executor — GH-5045
+//
+// Dispatch-time "base presence" guard: before a queued task's claim commits
+// to executing, check whether the issue body's own stated prerequisites
+// (an explicit "Depends on: #N" reference, or a backtick-quoted file path)
+// have actually landed on the target repo's default branch yet. A sub-issue
+// authored against a still-open sibling PR, or against a path a prior step
+// hasn't merged yet, has nothing to build on — executing it anyway wastes a
+// run reproducing work that's already in flight or racing a base that
+// doesn't exist.
+//
+// Mirrors issue_state.go's IssueStateChecker/fetchIssueState shape: a
+// narrow probe interface (BasePresenceProbe), a gh-CLI fallback
+// implementation, per-repo registration on Runner
+// (RegisterBasePresenceProbe/basePresenceProbeFor), and a swappable
+// package-level var (checkBasePresence) so dispatcher.go's claim-path check
+// and tests both call the exact same entry point.
+package executor
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+// resolveOwnerRepoForBasePresence resolves (owner, repo) for the claim-path
+// hold check, mirroring fetchIssueState's resolution order (issue_state.go):
+// task.SourceRepo when populated, else the `origin` remote at projectPath.
+// buildTaskFromExecution never populates SourceRepo, so the dispatcher
+// claim-path call site always falls through to the git-remote path.
+func resolveOwnerRepoForBasePresence(ctx context.Context, task *Task, projectPath string) (string, string, error) {
+	owner, repo := "", ""
+	if task.SourceRepo != "" {
+		owner, repo, _ = strings.Cut(task.SourceRepo, "/")
+	}
+	if owner != "" && repo != "" {
+		return owner, repo, nil
+	}
+	return resolveGitRemote(ctx, projectPath)
+}
+
+// labelPilotNeedsHuman is the escalation label applied once a held task has
+// exhausted DispatcherConfig.BasePresenceHoldMaxCycles held cycles without
+// its prerequisites landing — mirrors labelPilotSuperseded's
+// cycle-avoidance pattern (issue_state.go): defined once here rather than
+// imported from adapters/github to avoid the same import-cycle constraint
+// documented there.
+const labelPilotNeedsHuman = "pilot-needs-human"
+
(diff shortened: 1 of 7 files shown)


alekspetrov:
**Review verdict: REQUEST-CHANGES — DO NOT release from awaiting_approval as-is.**

The guard's plumbing is solid (pre-claim seam, ledger event + log asserted, zero-probe fast path with a call-count assertion, real incident-body fixtures), but two spec-level gaps need fixing before release. First, `basePresenceChecker.Check` (base_presence.go:126) only holds when the ref is itself an open PR; #5028 req 2a also requires holding when #N is an *issue* whose PR is open-unmerged — the exact `Depends on: #<sub-issue>` shape from the incident wave — and the current probe can't detect an attached PR at all. Second, the hold runs upstream of the GH-4656 issue-closed revalidation and always re-picks the same queue head: a persistent hold (e.g. genuinely-missing path) blocks the whole project queue indefinitely, escalation re-fires every 20 cycles after resetting its own counter, and even closing the issue can't unstick it since the superseded path is unreachable from a held row. Related: this PR defines its own probe interface with a signature incompatible with GH-5044's merged helpers and never registers an adapter probe, so subtask 1's code is dead and production always shells out to gh; and since all three tests stub `checkBasePresence` wholesale, the decision logic containing gap #1 has no direct coverage — a fake-probe unit test over `Check` would have caught it. Minor: the hold counter only resets on escalation, never on natural release, so stale counts cause premature escalation on later retries of the same key.

alekspetrov:
Closing unmerged per the REQUEST-CHANGES review above. The plumbing here (seam, observability, fixtures, fast path) was correct and the successor spec instructs reuse — but gaps F1 (issue-with-open-PR refs never hold) and F2 (persistent hold head-of-line-blocks the project queue with no exit) are release blockers, and the fresh-complete-spec route outperforms appended-scope retries (GH-5032 precedent: 1 of 4 appended items delivered). Superseded by #5050.

alekspetrov:
This PR was closed without merging: closed without merging (no reason recorded)

CI run: https://github.com/qf-studio/pilot/commit/9d45223f1ab3a9713a54148aa30d5956be0302da/checks

PR #5057, opened 2026-08-21
fix(executor): parse bare pilot-signal lines emitted outside fences

lkshrk:
## Summary

Residual of #4901, per the review there: everything the landed GH-4964 contract (#4977) covers has been dropped; this carries only the bare-signal parser fix.

The prompt mandates fenced `pilot-signal` blocks, but in production the model emits the payloads as bare `{"v":...}` lines (observed live on a watch-only ticket — trace in the #4901 thread). Those lines never match `signalBlockRegex`, so the whole signal surface — including the `no_op:true`+reason decline contract — is unreachable for that shape.

## Changes

- `ParseSignals` scans for bare signal lines after stripping fenced blocks, so nothing parses twice. Lines must start `{"v":` and parse as JSON; anything else is skipped.
- `signal_test.go` coverage: bare exit/status/no_op parsing, fence dedup, invalid-JSON and non-signal-JSON skips, and `NoOpExitReason` returning the reason from an unfenced `no_op` signal.

## Testing

`go test ./internal/executor/...` green (executor, ghguard, workflow).

### DIFF
--- a/internal/executor/signal.go
+++ b/internal/executor/signal.go
@@ -79,6 +79,19 @@ func (p *SignalParser) ParseSignals(text string) []PilotSignal {
 		signals = append(signals, signal)
 	}
 
+	// Models also emit payloads as bare lines outside a fence; fenced blocks are removed first so nothing parses twice.
+	for _, line := range strings.Split(signalBlockRegex.ReplaceAllString(text, ""), "\n") {
+		trimmed := strings.TrimSpace(line)
+		if !strings.HasPrefix(trimmed, `{"v":`) || !strings.HasSuffix(trimmed, "}") {
+			continue
+		}
+		var signal PilotSignal
+		if err := json.Unmarshal([]byte(trimmed), &signal); err != nil {
+			continue
+		}
+		signals = append(signals, p.validateSignal(signal))
+	}
+
 	return signals
 }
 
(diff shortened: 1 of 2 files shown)


alekspetrov (approved):
**Review: APPROVE.** Exactly the residual scoped in the #4901 thread, delivered tight: +73/-0, two files.

- Correct layering: bare-line scan runs over fence-stripped text, so nothing parses twice — and the "fenced and bare both parse once each" test pins it.
- Fails in the safe direction: a line must start `{"v":`, end `}`, and JSON-parse — anything mangled is skipped (missed signal → existing fallback paths), never misread. Bare no_op carries the #4977 reason contract through `validateSignal`, and the mandatory-exit-is-never-a-no-op case is pinned.
- This closes the real production gap: the live trace in #4901 showed the model emitting bare payloads, which made the entire signal surface — including the decline contract you got landed — unreachable for that shape.

Two non-blocking notes: (1) bare signals append after all fenced ones regardless of text position, so a bare status before a fenced status would misorder "latest progress" — theoretical with today's emission shape, worth a comment if it ever bites; (2) any answer-text line that legitimately quotes a `{"v":` example (e.g. a question task about the signal format itself) now parses as a real signal — same pre-existing exposure as quoted fences, accepted.

CI green. Merging. With this, #4901 is fully dispositioned on your side — feel free to close it. Thanks for carving the residual out so cleanly.

PR #5084, opened 2026-08-21
GH-5079: feat(alerts): emit escalation alert + park under pilot-needs-human on ladder exhaustion

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5079.

Closes #5079

## Changes

Wire the exhaustion event: when the executor advances the ladder to pilot-failed-retry-exhausted (from subtask 1), fire an escalation-class alert that renders through the PR#5069 event.Error path and apply pilot-needs-human so the issue parks under the GH-5056 admission check. Also address the dead loop-breaker branch in cmd/pilot/handler_common.go:237 (fireLoopBreakerAlert currently sits on the has-terminal-completion skip that no longer fires post-#5070) — either remove the dead branch or reroute it to the new exhaustion hook, whichever keeps the dispatch loop alert-covered. Add tests asserting alert content renders and needs-human label is applied on the exhausted transition only.

PR #5090, opened 2026-08-21
GH-4890: feat: alerts: persist active-alert state so recovery survives a restart

### LINKED ISSUE #4886: Alerts: notify when an alerted condition recovers

### LINKED ISSUE #4890: Alerts: persist active-alert state so recovery survives a restart

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-4890.

Closes #4890

## Changes

GitHub Issue GH-4890: Alerts: persist active-alert state so recovery survives a restart

### Problem

Follow-up to #4886. Active alerts live only in the engine's map, so a condition that recovers while the daemon is down produces no resolution at all: the restarted process has no record anything fired, and the healthy event is correctly but unhelpfully silent. Restarting is the most likely thing an operator does right after fixing a credential — exactly when recovery reporting is least able to work.

### Proposed Solution

Follow the pending-approval precedent: a new `active_alerts` table, an optional `ActiveAlertStore` the engine writes through, and a rehydrate call on startup. `*memory.Store` satisfies the interface directly.

Only the fields needed to emit a resolution are stored — including the channels the original alert was delivered to, so a rehydrated resolution still reaches those channels rather than being re-filtered on its own info severity.

Persistence is best-effort and off the alerting path: a store failure is logged and the alert still fires, so a broken database degrades recovery reporting rather than suppressing alerts. The store is optional; engines constructed without one behave exactly as before.

Implemented and running in production, will PR after resolution-notifications (stacked on it).

### Alternatives Considered

Re-deriving active state from the alert history table — history has no notion of "still firing" and doesn't record delivery channels.

### Affected Adapters

All (alerting core)

### Priority

Medium

---

## Maintainer addendum (2026-08-21, pre-dispatch)

Stacking blocker resolved: #4886 closed via PR#4898 (merged 08-17). Design as proposed is accepted — pending-approval persistence precedent, optional store, best-effort writes.

## Acceptance Criteria

- [ ] `ActiveAlertStore` optional interface in `internal/alerts`; `*memory.Store` satisfies it (compile-time assertion). Engines constructed without a store behave byte-identically (regression-pinned).
- [ ] New `active_alerts` table via the established duplicate-tolerant migration idiom (`internal/memory`); rows store only what a resolution needs, INCLUDING the channels the original alert was delivered to.
- [ ] Writes are best-effort and off the alerting path: store failure → log + alert still fires (test with a failing store).
- [ ] Rehydrate on startup; a condition that recovered while the daemon was down emits its resolution to the ORIGINALLY-delivered channels, not re-filtered by severity (real-SQLite e2e test: fire → persist → new engine instance → rehydrate → resolve → assert channel set).
- [ ] Firing/resolution lifecycle keeps the table bounded (resolved rows deleted or marked; no unbounded growth — state the choice).
- [ ] `go test ./internal/alerts/... ./internal/memory/...` green.

**Note**: the reporter states an implementation exists and runs in production on their fork. Per the PR#4900 precedent: if that PR arrives green before our executor's PR merges, theirs wins the overlap and ours rebases to residuals.

## Planned Steps (execute all in sequence)

- Step 1: **feat(memory): add active_alerts table and Store CRUD for persisted active-alert rows** — Add a duplicate-tolerant migration in `internal/memory` creating an `active_alerts` table with only the columns a resolution needs (alert key/id, condition metadata, severity, first-fired-at, and the delivered-channel set — serialized as JSON or a joined string, state the choice in the migration). Implement `*memory.Store` methods to Upsert an active alert on fire, Delete on resolve (bounded via delete-on-resolve; state that choice), and List/Load all currently-active rows for rehydrate. Cover the store methods with `internal/memory` unit tests (insert/list/delete round-trip, channel set preservation, migration idempotency). No changes outside `internal/memory` — the interface lives in `internal/alerts` and gets wired in the next subtask.
- Step 2: **feat(alerts): persist active alerts through optional ActiveAlertStore and rehydrate resolutions on startup** — In `internal/alerts`, define the optional `ActiveAlertStore` interface (Upsert / Delete / List) and add a compile-time assertion that `*memory.Store` satisfies it. Wire the engine to write through the store on fire and delete on resolve, strictly best-effort and off the alerting path — a store error is logged and the alert still fires/resolves (test with a deliberately-failing store double). Add a Rehydrate/Load call invoked at engine construction that pulls persisted rows and re-seeds the in-memory active-alerts map, tagging entries with their originally-delivered channel set so a post-restart recovery emits its resolution to those exact channels rather than being re-filtered by info severity. Regression-pin that an engine constructed without a store behaves byte-identically to today. Add a real-SQLite end-to-end test in this package (fire → persist → construct a fresh engine against the same DB → rehydrate → trigger recovery → assert the resolution reached the original channel set and the row was deleted). Verify `go test ./internal/alerts/... ./internal/memory/...` is green.


alekspetrov:
**Post-merge review — APPROVE-w-DEFECT: implementation and tests check out against every addendum AC, but the store is NEVER WIRED — the feature is dead code as merged** (verdict as comment — own-account PR; verified at `a7d5141a`, all three suites green).

**All six ACs verified:**
- Interface + compile-time assertion (`active_alert_store.go:14-21`); nil-store byte-identical behavior pinned.
- `active_alerts` table via the migrations-slice idiom, `IF NOT EXISTS` idempotent (reopen pinned); channels JSON round-trip pinned.
- Best-effort confirmed with an all-methods-failing store; persist runs after the alert fires, outside the mutex.
- **The rehydrate e2e is well-constructed**: the test channel's `Severities` excludes `info`, so severity re-filtering would fail the test — and the code path is genuinely direct (`dispatchResolution` → `enqueueDispatch(active.channels)` → `Dispatcher.Dispatch`, which does no severity check for explicit channel lists).
- Boundedness: delete-on-resolve, stated in the migration comment.

**The defect — zero production call sites for `WithActiveAlertStore`.** All three engine-construction sites have a `*memory.Store` in scope and already append store-dependent options: `cmd/pilot/main.go:3082` (polling mode, inside the existing `if store != nil` block), `cmd/pilot/main.go:1040` (gateway), `internal/pilot/pilot.go:1659`. Until those one-liners land, a real restart still loses all active-alert state — GH-4716 class, and FEATURE-MATRIX already marks the row ✅. Follow-up filed + dispatched.

Non-blocking observations: (1) cooldowns aren't rehydrated — pre-existing, but the persisted `created_at` could seed `lastAlertTimes` cheaply; (2) #5084's exhaustion escalations correctly don't persist (no resolution lifecycle); persistence scope intentionally equals the #4886 resolution scope.

**@lkshrk courtesy notes** (for diffing against your fork): rehydrate is implicit inside `NewEngine` (unexported) rather than an explicit startup call; persistence is gated on the resolve-notifications toggle; schema is PK `(rule_name, source)`, JSON `metadata`/`channels` TEXT columns, no `resolved_at` (delete-on-resolve). Interface/methods/table name match your issue text — a fork diff should be mostly mechanical.

PR #5120, opened 2026-08-21
feat(telegram): reply in the forum topic a message arrived from

### LINKED ISSUE #4887: Telegram: replies should return to the forum topic a message arrived from

lkshrk:
Inbound leg of the forum-topic chain (#4871 outbound merged via #4891).

Captures `message_thread_id` from incoming long-poll updates, gated on `is_topic_message` so supergroup reply chains are not mistaken for topics, and forwards it as `ThreadID` on the `IncomingMessage` so `comms.Handler` threads confirmations, results and chunked output back to the originating topic. General-thread messages carry no topic and keep the configured default.

Fixes #4887

Rebased onto current main (the old copy of the #4891 commit dropped). Single commit; full rationale in the commit message. Verified: `go build/vet/test ./...` clean, `golangci-lint run --new-from-rev=main` 0 issues.

Stacked follow-up for #4888 (SendText threading) comes next; it includes this commit.

### DIFF
--- a/docs/content/integrations/telegram.mdx
+++ b/docs/content/integrations/telegram.mdx
@@ -92,6 +92,18 @@ Approval prompts addressed to an individual approver (where `approvers[0]` is a
 chat ID) go to that chat directly and ignore this setting. Alert channels take
 their own `message_thread_id` — see [Alerts](/features/alerts).
 
+### Replies follow the conversation
+
+Replies to a message you sent — answers, plans, task confirmations, results —
+go back to the topic that message came from, whichever topic that is.
+`message_thread_id` is the default for messages Pilot starts on its own (task
+lifecycle notifications and alerts), not a ceiling on where replies can go.
+
+Messages sent in a forum's General thread are answered in General. In a
+supergroup *without* Topics enabled, replies are unaffected — the Bot API only
+accepts a topic for forum supergroups, so Pilot deliberately ignores the thread
+id of an ordinary reply chain rather than sending one that would be rejected.
+
 ## Chat Modes
 
 Pilot automatically detects your intent and responds in one of five modes:
--- a/internal/adapters/telegram/client.go
+++ b/internal/adapters/telegram/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"time"
 )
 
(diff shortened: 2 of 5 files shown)


alekspetrov (approved):
## Review — inbound leg of the forum-topic chain (#4887)

**Verdict: APPROVE (with two follow-up notes).**

Inbound leg looks right. Verified the `is_topic_message` gate against the Bot API semantics (reply-chain `message_thread_id` correctly ignored; General thread falls through to no-param), and the wire path end to end: `topicThreadID()` → `IncomingMessage.ThreadID` → `comms.Handler` → `parseThreadID("")` → `0` → `omitempty`, so non-topic behavior is byte-identical to before. Diff is tightly scoped (5 files, single rebased commit as described — the #4891-class stale-fork sweep did not recur), CI green across all 7 gates including Argument-Discarding Mocks (the `threadRecordingMessenger` records and asserts the threadID — appreciated), and the table-driven tests pin all three gate cases at both the client and handler-integration level. Callback construction sites covered too.

Two non-blocking gaps to track (out of #4887's stated scope, but they slightly overstate the new docs' "whichever topic that is" promise):
- `handleVoice` builds its `IncomingMessage` without `ThreadID` (handler.go:631) — a voice message sent in a topic gets its transcript and reply outside the topic.
- `handlePhoto` calls `ExecuteDirectTask(ctx, chatID, "", ...)` with a hardcoded empty threadID (handler.go:550).

Filed as a follow-up issue (linked below) — feel free to pick it up if you want to complete the chain, otherwise it queues for the daemon. Merging now; #5121 review is running in parallel and can rebase to its single SendText commit after this lands.

PR #5166, opened 2026-08-24
GH-5155: feat(telegram): add pre-mutation authorization guard to HandleCallback

### LINKED ISSUE #5155: feat(telegram): add pre-mutation authorization guard to HandleCallback
Parent: GH-5153

In internal/approval/telegram.go HandleCallback (~:471), add identity checks before any state mutation, mirroring the existing not-found/expired early-return shape. Two-step guard: (1) if pending.Request.Approvers is non-empty, require plain string equality of userID against an entry; (2) otherwise fall back to a handler-scoped allowlist. Empty approvers + empty fallback = unrestricted.

## Scope fence

Implement ONLY the slice described above. The parent issue GH-5153 is decomposed

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5155.

Closes #5155

## Changes

In internal/approval/telegram.go HandleCallback (~:471), add identity checks before any state mutation, mirroring the existing not-found/expired early-return shape. Two-step guard: (1) if pending.Request.Approvers is non-empty, require plain string equality of userID against an entry; (2) otherwise fall back to a handler-scoped allowlist. Empty approvers + empty fallback = unrestricted.

### DIFF
--- a/.agent/tasks/gh-5155.md
+++ b/.agent/tasks/gh-5155.md
@@ -1,0 +1,10 @@
+# GH-5155
+
+**Created:** 2026-08-24
+
+## Problem
+
+In internal/approval/telegram.go HandleCallback (~:471), add identity checks before any state mutation, mirroring the existing not-found/expired early-return shape. Two-step guard: (1) if pending.Request.Approvers is non-empty, require plain string equality of userID against an entry; (2) otherwise fall back to a handler-scoped allowlist. Empty approvers + empty fallback = unrestricted.
+
+## Acceptance Criteria
+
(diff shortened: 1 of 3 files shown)


alekspetrov:
Post-merge review: **APPROVE**. Authorization decided inside the same critical section as the pending-map lookup; unauthorized taps mutate nothing (pending retained, store untouched, no response emitted) and get a visible refusal via AnswerCallback; listed approver can still decide afterward. Tests cover intruder-then-approver, allowlist fallback, and the routing fixture update. Notes: builder landed as `WithAllowedUsers` (spec said `WithAllowedIDs`) — better name, but now diverges from Slack's `WithAllowedIDs` in #5169; refusal logs at Info (spec said Warn) — fine.

Issue #5063, opened 2026-08-21
Root repo core.bare flips to true when SIGTERM kills a pre-push gate mid-run (3rd occurrence; writer still unidentified)


alekspetrov:
## Summary

`.git/config`'s `core.bare` has flipped from `false` to `true` three times (2026-07-28, 2026-08-18, 2026-08-21), always correlated with a `git push` being SIGTERM'd (2-min timeout) while `scripts/pre-push-gate.sh`'s `go test -short -race ./...` step (measured 188-238s, see `.agent/sops/quality/pre-push-gate.md`) was still running. Because worktrees share `.git/config`, the flip breaks `git status`/build/lint/test for EVERY concurrent session (root + all linked worktrees), not just the pusher — repo-wide, not cosmetic.

## What's confirmed

- No script/hook in this repo (`scripts/*.sh`, `.git/hooks/pre-push`) or Go code (`internal/`, `cmd/`) writes `core.bare`, `--bare`, `GIT_DIR`, or `--work-tree` — verified by grep across three separate investigations.
- No trap/cleanup exists anywhere in the pre-push scripts (grep for `trap ` returns nothing), so there's no restore-on-EXIT-but-not-TERM asymmetry — nothing ever unwinds `core.bare` because nothing sets it locally.
- The measured full-gate runtime (224-287s) exceeds a 120s SIGTERM window by construction, so the hang is expected/reproducible, not a network stall.
- The actual writer of `bare=true` remains **unidentified** after 3 incidents.

## Ask

1. Raise the timeout on any automation that `git push`es against this repo to ≥300s, or route non-interactive pushes through `--no-verify` + CI-side gating (CI runs the same gates).
2. Add a defensive pre-flight (`git rev-parse --is-bare-repository`) to session-start tooling (pilot-board / nav-start hook) that auto-corrects and logs a `core.bare=true` drift instead of requiring manual triage each occurrence.
3. Instrument the next reproduction with `GIT_TRACE=1`/`GIT_TRACE_SETUP=1` or a `.git/config` mtime watch to actually catch the writer — static code review has been exhausted across 3 attempts.

## Note

⚠️ Touches the local quality-gate surface — left unlabeled for founder review per policy; label `pilot` to dispatch asks 1–2 if approved (ask 3 is operator instrumentation).

## References

- `.agent/knowledge/memories/pitfalls/core-bare-true-poisons-status-and-vcs-stamping.md` (07-28, updated 08-08 + today)
[...]

PR #5222, opened 2026-08-25
GH-5218: chore(scripts): package git-config-watch as an armable service (make target + systemd unit + launchd plist, single-instance) so occurrence №5 isn't missed

### LINKED ISSUE #5218: chore(scripts): package git-config-watch as an armable service (make target + systemd unit + launchd plist, single-instance) so occurrence №5 isn't missed
## Context

`scripts/git-config-watch.sh` (GH-5063, merged in PR#5208) is the forensic tripwire for the still-unidentified writer that has flipped the shared git config's core.bare false→true four times (2026-07-28, 08-18, 08-21, 08-24), breaking every concurrent session at once. By design it is passive operator-opt-in tooling — and in practice that means nothing runs it: occurrence №5 will be missed unless an operator happens to have a tmux pane open. The script's own header documents the ad-hoc tmux recipe for both the laptop and the founder box, but there is no packaged way to keep it armed.

Known negative (2026-08-24): SIGTERM'ing the pre-push gate mid-test did NOT reproduce the flip — the watcher must be armed continuously around push activity, not only while the gate runs, and must not be wired INTO the gate (the gate is itself a suspect).

## Task

Do not decompose this issue — it is a single-subsystem change that must land as one PR.

Package the watcher so arming it is one command per machine, still operator-initiated, never automatic:

1. Makefile target watch-git-config that execs `scripts/git-config-watch.sh` in the foreground (thin wrapper, inherits GCW_* env vars).
2. A systemd user-unit file, committed as a new file under scripts/ (suggested name git-config-watch.service, alongside the script), with Restart=on-failure and the forensic log at a stable path, plus install/enable instructions in its comment header — for the founder box, where pushes come from autopilot and no tmux pane survives. The unit is shipped disabled; enabling it is an operator step.
3. A launchd plist equivalent for macOS, also committed under scripts/ (suggested name com.pilot.git-config-watch.plist), same shipped-disabled contract — for the laptop push context.
4. Single-instance guard in the script itself (lockfile or pgrep check): a second launch must refuse and point at the running instance, so the unit + a stray manual tmux copy never double-heal or interleave forensics.
5. Update the script's header How-to-run section to document all three paths (make target, systemd, launchd) and state the shipped-disabled contract.

No cron, no CI wiring, no automatic start from any daemon code path — this stays operator-armed. Shell only; no Go changes.

## Acceptance

- [ ] make watch-git-config runs the watcher from the repo root and from a linked worktree.
- [ ] Second concurrent launch exits non-zero with a message naming the running instance (test via the script's own test harness style used by Box Scripts CI — ShellCheck must pass).
- [ ] Unit/plist files pass a syntax check (systemd-analyze verify is not available on macOS CI — a comment-documented manual step is acceptable; ShellCheck covers the script changes).
- [ ] Script header documents all run modes; shipped-disabled contract stated in both unit files.

## Refs

- GH-5063 · PR#5208 (script + 4th-occurrence datapoint + SIGTERM negative)
- `scripts/git-config-watch.sh` (header documents current ad-hoc tmux recipes)
- `scripts/install-hooks.sh` (existing operator-tooling install pattern)

alekspetrov:
## Summary

Automated PR created by Pilot for task GH-5218.

Closes #5218

## Changes

GitHub Issue GH-5218: chore(scripts): package git-config-watch as an armable service (make target + systemd unit + launchd plist, single-instance) so occurrence №5 isn't missed

## Context

`scripts/git-config-watch.sh` (GH-5063, merged in PR#5208) is the forensic tripwire for the still-unidentified writer that has flipped the shared git config's core.bare false→true four times (2026-07-28, 08-18, 08-21, 08-24), breaking every concurrent session at once. By design it is passive operator-opt-in tooling — and in practice that means nothing runs it: occurrence №5 will be missed unless an operator happens to have a tmux pane open. The script's own header documents the ad-hoc tmux recipe for both the laptop and the founder box, but there is no packaged way to keep it armed.

Known negative (2026-08-24): SIGTERM'ing the pre-push gate mid-test did NOT reproduce the flip — the watcher must be armed continuously around push activity, not only while the gate runs, and must not be wired INTO the gate (the gate is itself a suspect).

## Task

Do not decompose this issue — it is a single-subsystem change that must land as one PR.

Package the watcher so arming it is one command per machine, still operator-initiated, never automatic:

1. Makefile target watch-git-config that execs `scripts/git-config-watch.sh` in the foreground (thin wrapper, inherits GCW_* env vars).
2. A systemd user-unit file, committed as a new file under scripts/ (suggested name git-config-watch.service, alongside the script), with Restart=on-failure and the forensic log at a stable path, plus install/enable instructions in its comment header — for the founder box, where pushes come from autopilot and no tmux pane survives. The unit is shipped disabled; enabling it is an operator step.
3. A launchd plist equivalent for macOS, also committed under scripts/ (suggested name com.pilot.git-config-watch.plist), same shipped-disabled contract — for the laptop push context.
4. Single-instance guard in the script itself (lockfile or pgrep check): a second launch must refuse and point at the running instance, so the unit + a stray manual tmux copy never double-heal or interleave forensics.
5. Update the script's header How-to-run section to document all three paths (make target, systemd, launchd) and state the shipped-disabled contract.

No cron, no CI wiring, no automatic start from any daemon code path — this stays operator-armed. Shell only; no Go changes.

## Acceptance

- [ ] make watch-git-config runs the watcher from the repo root and from a linked worktree.
- [ ] Second concurrent launch exits non-zero with a message naming the running instance (test via the script's own test harness style used by Box Scripts CI — ShellCheck must pass).
- [ ] Unit/plist files pass a syntax check (systemd-analyze verify is not available on macOS CI — a comment-documented manual step is acceptable; ShellCheck covers the script changes).
- [ ] Script header documents all run modes; shipped-disabled contract stated in both unit files.

## Refs

- GH-5063 · PR#5208 (script + 4th-occurrence datapoint + SIGTERM negative)
- `scripts/git-config-watch.sh` (header documents current ad-hoc tmux recipes)
- `scripts/install-hooks.sh` (existing operator-tooling install pattern)


alekspetrov:
## Post-merge review — APPROVE with notes (one gap: test not wired into CI)

**Verdict: APPROVE.** All four acceptance criteria functionally met; full lock-guard test suite run locally at the merge commit (`67543ed2`): **8/8 pass**, ShellCheck clean on both scripts. The shipped-disabled contract is stated in all three places (script header, unit, plist) and is real — RunAtLoad=false in the plist, no Install-time start in the unit, no daemon/CI wiring anywhere.

### Notes

1. **Gap — the new test suite never runs in CI.** The Box Scripts job invokes `scripts/box/pilot-backup-s3_test.sh` by explicit path (ci.yml:70) and this PR didn't add `scripts/git-config-watch_test.sh` alongside it — the spec's acceptance bullet asked for the harness style, and the style was matched, but regression protection is currently manual. One-line CI fix; folding it into the next CI-touching PR is fine.
2. **Lock acquire has a check-then-write race** (`kill -0` probe, then `echo $$ >`) rather than an atomic `mkdir`/`flock` take — two truly simultaneous launches can both pass the probe and the second silently overwrites the first's lock. For human/service launch cadence the window is negligible, and the failure mode (two watchers) is exactly what an atomic lock would prevent, so noting rather than requesting change. The stale-reclaim path makes an orphaned lock self-healing, which the tests prove.
3. **Lock-release-on-SIGTERM works** (verified by the passing release test — the EXIT trap fires on TERM here), and even where it wouldn't, note 2's stale reclaim covers it.
4. The systemd unit hardcodes the box checkout path in ExecStart with an edit note, and the plist ships a `/path/to/your/...` placeholder — both documented as operator-edited. The `loginctl enable-linger` line in the unit header is the load-bearing instruction for the box (no persistent login session); good catch by the implementation.

### Operator steps now open (not autopilotable)

- Box: copy unit → `systemctl --user enable --now` → `loginctl enable-linger`.
- Laptop: copy plist, edit path → `launchctl load` + `start`.

Once armed on both, the occurrence-№5 hunt is finally continuous rather than tmux-pane-dependent.

